Issue #18, opened 2021-01-14
AWS region for deployments service is fixed to us-east-1


fabianmp:
The chart does not provide a method to specify the AWS region, which defaults to us-east-1.
This causes an error in our deployment because our Kubernetes cluster is running in a different region.

Since the region can be overridden for the deployments service by setting the environment variable DEPLOYMENTS_AWS_REGION I suggest to add this option similar to the other S3 settings (e.g. AWS_BUCKET).

tranchitella:
Thanks for your report @fabianmp, I'll take care of it.

tranchitella:
Merged to master

PR #19, opened 2021-01-14
Add the global.s3.AWS_REGION setting (#18)

tranchitella:
Changelog: title


lluiscampos (approved):
LGTM.

The pipeline error seems unrelated, but we need to fix somehow.

PR #23, opened 2021-03-11
MEN-4486: Update chart for Mender 2.6.0

lluiscampos:
Adding new services auditlogs and deviceconnect.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,14 +8,14 @@ Using `helm3`:
 
 ```bash
 $ make package
-$ helm install mender ./mender-2.5.0.tgz
+$ helm install mender ./mender-2.6.0.tgz
 ```
 
 or using `helm2`:
 
 ```bash
 $ make package
-$ helm install --name mender ./mender-2.5.0.tgz
+$ helm install --name mender ./mender-2.6.0.tgz
 ```
 
 ## Introduction
@@ -32,13 +32,13 @@ This chart bootstraps a [Mender](https://mender.io) deployment on a [Kubernetes]
 To install the chart with the release name `my-release` using `helm3`:
 
 ```bash
-$ helm install my-release -f values.yaml ./mender-2.5.0.tgz
+$ helm install my-release -f values.yaml ./mender-2.6.0.tgz
 ```
 
 or using `helm2`:
 
 ```bash
-$ helm install --name my-release -f values.yaml ./mender-2.5.0.tgz
+$ helm install --name my-release -f values.yaml ./mender-2.6.0.tgz
 ```
 
 The command deploys Mender on the Kubernetes cluster in the default configuration. The [Parameters](#parameters) section lists the parameters that can be configured during installation.
@@ -138,13 +138,13 @@ Specify each parameter using the `--set key=value[,key=value]` argument to `helm
 ```bash
 $ helm install my-release \
   --set mongodbRootPassword=secretpassword,mongodbUsername=my-user,mongodbPassword=my-password,mongodbDatabase=my-database \
-  ./mender-2.5.0.tgz
+  ./mender-2.6.0.tgz
 ```
 
 Alternatively, a YAML file that specifies the values for the parameters can be provided while installing the chart. For example,
 
 ```bash
-$ helm install --name my-release -f values.yaml ./mender-2.5.0.tgz
+$ helm install --name my-release -f values.yaml ./mender-2.6.0.tgz
 ```
 
 > **Tip**: You can use the default [values.yaml](values.yaml)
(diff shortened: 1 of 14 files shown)


mender/templates/gui-deploy.yaml
```diff
@@ -77,4 +77,8 @@ spec:
           value: {{ .Values.gui.env.GATEWAY_IP | quote }}
         - name: GATEWAY_PORT
           value: {{ .Values.gui.env.GATEWAY_PORT | quote }}
+        {{- if .Values.global.deviceconnect }}
+        - name: HAVE_DEVICECONNECT
+          value: {{ .Values.global.deviceconnect | quote }}
+        {{- end }}
```
lluiscampos, line 83: For these (both here and in workflows-worker) I don't know if I should use `Values.deviceconnect.enabled` instead of the new global that I have declared

README.md
```diff
@@ -440,6 +495,10 @@ $ helm repo update
 $ helm install --name minio --version 6.0.5 --set "accessKey=myaccesskey,secretKey=mysecretkey" minio/minio
 ```
 
+### Installing NATS
+
+TODO
```
lluiscampos, line 500: What should I do here? If we recommend a helm chart to install we should use it too for testing.

lluiscampos:
@tranchitella Please review carefully the templates ;)

mender/templates/workflows-worker-deploy.yaml
```diff
@@ -52,6 +52,16 @@ spec:
         args: ["worker", "--excluded-workflows", "generate_artifact"]
 {{- end }}
 
+        env:
+        {{- if .Values.global.auditlogs }}
+        - name: HAVE_AUDITLOGS
+          value: "1"
```
lluiscampos, line 58: Here I cannot use `Values.global.auditlogs` because it would be "true" while the service expects "1", correct?

tranchitella: Any value is good, 1 or true, doesn't matter. Just use the variable :)

README.md
```diff
@@ -157,14 +159,14 @@ The following table lists the parameters for the `api-gateway` component and the
...
 | `api_gateway.image.registry` | Docker image registry | `docker.io` |
 | `api_gateway.image.repository` | Docker image repository | `mendersoftware/api-gateway` |
-| `api_gateway.image.tag` | Docker image tag | `mender-2.5.0` |
+| `api_gateway.image.tag` | Docker image tag | `mender-2.6.0` |
 | `api_gateway.image.imagePullPolicy` | Docker image pull policy | `IfNotPresent` |
 | `api_gateway.replicas` | Number of replicas | `1` |
 | `api_gateway.affinity` | [Affinity map](https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#affinity-and-anti-affinity) for the POD | `{}` |
 | `api_gateway.resources.limits.cpu` | Resources CPU limit | `500m` |
-| `api_gateway.resources.limits.cpu` | Resources memory limit | `128M` |
+| `api_gateway.resources.limits.memory` | Resources memory limit | `128M` |
```
tranchitella, line 167: Thanks for fixing these typos!

README.md
tranchitella: You can simply link to here: https://nats-io.github.io/k8s
I don't think we need instruct the user how to install the Helm chart (I'm also thinking about removing these details for MongoDB and Minio, TBH). By the way, nats is an extremely simple service to run, no persistence, only a container running. I'm okay in testing it as we are doing now, with yamls applied with kubectl apply.

tests/nats/configmap-nats-config.yaml
```diff
@@ -0,0 +1,24 @@
...
+  nats.conf: |
+    pid_file: "/var/run/nats/nats.pid"
+    http: 8222
+
+    cluster {
+      port: 6222
+      routes [
+        nats://nats-0.nats.prod.svc:6222
+        nats://nats-1.nats.prod.svc:6222
+        nats://nats-2.nats.prod.svc:6222
```
tranchitella, line 15: As you are deploying nats with replicas set to 1, you'll only get the first one (nats-0).
By the way, the URL embeds the namespace, which is wrong in this context. Let's stick to namespace-local domain names, and use a single server.

```suggestion
        nats://nats-0.nats:6222
```

tests/nats/mender-nats.yaml
```diff
@@ -0,0 +1,108 @@
...
+        - name: POD_NAME
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.name
+        - name: POD_NAMESPACE
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.namespace
+        - name: CLUSTER_ADVERTISE
+          value: $(POD_NAME).nats.$(POD_NAMESPACE).svc
```
tranchitella, line 70: Counterpart of the configmap changee:

```suggestion
          value: $(POD_NAME).nats
```

tranchitella (approved):
Looks good; I have a couple of comments, but all in all you've done a great job!

mender/templates/gui-deploy.yaml
tranchitella: Both approaches are good; to be honest, I would have not introduced the "globals" and only used Values.deviceconnect.enabled. There is no point in having the deviceconnect service but not enabling it in the UI and in workflows...

lluiscampos: Makes sense. I originally just followed what was done for multitenancy, and then question it. I'll clean it up!

tranchitella (approved):

PR #25, opened 2021-03-16
Pass SSL env variable to api-gateway

nucatus:
add a SSL environment variable frontend for the api-gateway container SSL environment variable.


lluiscampos (changes requested):
Thanks @nucatus for your contribution. Two comments:

* Please add the new `SSL` value in the corresponding table in `README`.
* Formality: you need to sign your commit. See [Contributing](https://github.com/mendersoftware/mender/blob/master/CONTRIBUTING.md#sign-your-work) (I just realized it is missing in this repo, so I'll fix that independently)

Other than that looks good to me.

lluiscampos (approved):

tranchitella (approved):

PR #26, opened 2021-03-18
Remove SSL port mapping svc->pod if SSL disabled

nucatus:
If helm chart is deployed without SSL support, the api-gateway-svc mapping
of the SSL port should also be removed.

Signed-off-by: Alexandru Ionita <alexandru.ionita@gmail.com>

### DIFF
--- a/mender/templates/api-gateway-svc.yaml
+++ b/mender/templates/api-gateway-svc.yaml
@@ -29,6 +29,7 @@ spec:
   loadBalancerSourceRanges: {{- toYaml .Values.api_gateway.service.loadBalancerSourceRanges | nindent 4 }}
   {{- end }}
   ports:
+{{- if .Values.api_gateway.env.SSL }}
   - port: {{ .Values.api_gateway.service.httpsPort }}
     protocol: TCP
     targetPort: 443
@@ -36,6 +37,7 @@ spec:
     nodePort: {{ .Values.api_gateway.service.httpsNodePort }}
     {{- end }}
     name: https
+{{- end }}
   - port: {{ .Values.api_gateway.service.httpPort }}
     protocol: TCP
     targetPort: 80


mender/templates/api-gateway-svc.yaml
```diff
@@ -29,13 +29,15 @@ spec:
   loadBalancerSourceRanges: {{- toYaml .Values.api_gateway.service.loadBalancerSourceRanges | nindent 4 }}
   {{- end }}
   ports:
+{{- if .Values.api_gateway.env.SSL }}
```
lluiscampos, line 32: Shouldn't this be:
```suggestion
{{- if .Values.api_gateway.service.SSL }}
```
?

nucatus: According to the current context, the SSL flag is defined in the env scope, thus the `api_gateway.env.SSL` variable name is correct.

However, changing the scope of that flag would be appropriate, since it affects the entire service and it is not localized to the environment of a particular container, although it has implications to such environment. 

In this case I suggest to rename the flag to something like this: `api_gateway.service.ssl_enabled` and group it together with the variables `api_gateway.service.httpsPort` and `api_gateway.service.httpsNodePort`.

Maybe @tranchitella or somebody else more familiarized with this context can bring more insight into how to proceed further.

Thanks.

nucatus:
any feedback on this one? It feels like the initial fix is not complete and it would be nice to have the change consistent across the system.

lluiscampos:
@tranchitella Can you please have a look? I don't know what the best practice is for this.

mender/templates/api-gateway-svc.yaml
tranchitella: @nucatus you are right, we should probably move the flat to the service. At this stage, though, keeping it as it is is the easiest choice and doesn't introduce a back-compatibility breaking change.

tranchitella (approved):

PR #29, opened 2021-06-08
support new production install path via helm

mchalski:
https://tracker.mender.io/browse/MEN-4679

For now, it's a half done solution, with known issues, to be discussed more and tuned perhaps.
Quite a lot of it works though, the helm charts are actually very comprehensive.

What's inside:

**Templates/examples for helm value files**
This is supposed to be the main artifact supporting a production install. 

We prepare a set of 'values' files which the user can fill in with their certs, envs, etc. and simply do `helm install` on them.
We also give them updated examples of already filled in templates, so using those as an example, the install is simply:
```
make chart
helm install mender -f examples/example.values.enterprise.yaml mender-2.7.0.tgz
```
I wanted to fragment those files more, similar to the demo setup (os, enterprise, storage, storage-minio, etc.), but merging multiple files is risky with helm - the last file wins and it's very easy to overwrite a shared section.
I went with file-per-env, but is it a viable solution? there will be many more combinations of components and settings

**Tweaks to helm minio handling** 
I added a missing minio config to traefik configmap so we can route from outside to it (important for the client, deployments).
Minio can also now be exposed on the same domain as the gateway - no need for users to prepare another domain, which was problematic for them.
We use a port to discriminate api vs storage access.

**Misc tweaks to (root) cert handling**
This part is naively implemented and confuses me actually. It was needed to progress though and I didn't get back to it.

First of all, if we use self signed certs, they must not be 0 depth anymore. there must be even a bogus root CA.
Deployments (and client btw) insists that it must have the rootCA in its trust store; there is no work around to this check.

I defined a separate global field with the root CA, to be used in a new deployments secret. It mounts under the STORAGE_BACKEND_CERT env.

To avoid duplication, it should probably be picked from the gateway's cert (which must already bundle the cert and the root CA cert), and mounted as the same secret. 

**Known issues**
- artifact generation not working


tranchitella:
Superseded

PR #31, opened 2021-06-15
Handle no SSL termination in the api-gateway

nucatus:
In case no SSL termination is desired in the api-gateway, that should be
possible and the traefik configuration should be consistent with the
supplied configuration parameter `api_gateway.env.SSL`.

Signed-off-by: Alexandru Ionita <alexandru.ionita@gmail.com>

### DIFF
--- a/mender/templates/api-gateway-configmap.yaml
+++ b/mender/templates/api-gateway-configmap.yaml
@@ -1,3 +1,5 @@
+{{ $entrypoint := ternary "https" "http" (.Values.api_gateway.env.SSL) }}
+{{ $isTls := ternary true false (.Values.api_gateway.env.SSL) }}
 apiVersion: v1
 kind: ConfigMap
 metadata:
@@ -23,7 +25,7 @@ data:
         # auditlogs
         #
         auditlogs:
-          entrypoints: https
+          entrypoints: {{ $entrypoint }}
           middlewares:
           - userauth
           - sec-headers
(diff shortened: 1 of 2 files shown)


mender/templates/api-gateway-configmap.yaml
```diff
@@ -1,3 +1,5 @@
+{{ $entrypoint := ternary "https" "http" (.Values.api_gateway.env.SSL) }}
```
ffoysal, line 1:
```suggestion
{{ $scheme := ternary "https" "http" (.Values.api_gateway.env.SSL) }}
```
with this single variable might refer to everywhere.

mender/templates/api-gateway-configmap.yaml
```diff
@@ -417,10 +420,11 @@ data:
             regex: "^/api/management/v1/inventory/(.*)"
             replacement: "/api/0.1.0/$1"
 
+{{- if .Values.api_gateway.env.SSL }}
```
ffoysal, line 423:
```suggestion
{{- if $isTls }}
```

nucatus: Comments accepted and addressed in commit 53d7e06.

mender/templates/api-gateway-configmap.yaml
nucatus: Comments accepted and addressed in commit 53d7e06.

ffoysal (approved):

tranchitella (approved):
Looks good, thank you for your contribution

PR #32, opened 2021-06-16
Add Apache v2 LICENSE, CI check, and update README

lluiscampos:


tranchitella (approved):

lluiscampos:
Note to self: add also the checksum and the license check test before merge

PR #41, opened 2021-07-09
Downgrade mongodb Chart version to 10.21.1, as 10.21.2 is not released

tranchitella:
Changelog: none


Nicarim (approved):

tranchitella:
No idea what's up with GitLab, merging without the status report.

PR #49, opened 2021-08-04
Use `auditlogs.enabled` as flag to enable auditlogs across services

tranchitella:
Changelog: none

### DIFF
--- a/mender/templates/gui-deploy.yaml
+++ b/mender/templates/gui-deploy.yaml
@@ -79,16 +79,16 @@ spec:
         - name: HAVE_MULTITENANT
           value: {{ .Values.global.enterprise | quote }}
         {{- end }}
-        {{- if .Values.auditlogs.enabled }}
+        {{- if and .Values.auditlogs.enabled .Values.global.enterprise }}
         - name: HAVE_AUDITLOGS
-          value: {{ .Values.auditlogs.enabled | quote }}
+          value: "true"
         {{- end }}
         {{- if .Values.deviceconnect.enabled }}
         - name: HAVE_DEVICECONNECT
-          value: {{ .Values.deviceconnect.enabled | quote }}
+          value: "true"
         {{- end }}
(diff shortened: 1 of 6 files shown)


merlin-northern (approved):
lgtm

README.md
```diff
@@ -138,6 +138,7 @@ The following table lists the global parameters supported by the chart and their
 | Parameter | Description | Default |
 | -------------------------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------- | -------------------------------------------------------- |
 | `global.enterprise` | Enable the enterprise features | `true` |
+| `global.auditlogs` | Enabled the auditlogs feature (requires enterprise) | `true` |
```
merlin-northern, line 141: I wonder gui uses `.Values.auditlogs.enabled`

merlin-northern: andso does workflows-worker

tranchitella: You are right, doesn't make sense to have both flags.
I reworked the PR, let's use `auditlogs.enabled` everywhere.

mender/templates/gui-deploy.yaml
```diff
@@ -79,7 +79,7 @@ spec:
         - name: HAVE_MULTITENANT
           value: {{ .Values.global.enterprise | quote }}
         {{- end }}
-        {{- if .Values.auditlogs.enabled }}
+        {{- if and .Values.auditlogs.enabled .Values.global.enterprise }}
         - name: HAVE_AUDITLOGS
           value: {{ .Values.auditlogs.enabled | quote }}
```
bitcloud, line 84: This variable doesn't make sense. If it is false then the whole block doesn't exist. so why not do `value: "true"` as in all the other cases?

tranchitella: @bitcloud you are right, I force-pushed setting the value to `"true"`

merlin-northern (approved):
looks good to me

PR #51, opened 2021-09-15
Update the Helm Chart for Mender 3.1.0

tranchitella:


lluiscampos (approved):
Looks good! I couldn't spot any error. Let's wait to have the final tags, replace in README, and then merge :+1:

PR #56, opened 2021-10-18
Added support for deviceauth enterprise in Helm Chart

MaciejTe:
Added support for deviceauth enterprise in Helm Chart

Changelog: None

Signed-off-by: Maciej Tomczuk <maciej.tomczuk@northern.tech>

### DIFF
--- a/mender/values.yaml
+++ b/mender/values.yaml
@@ -95,8 +95,8 @@ device_auth:
       memory: 128M
   affinity: {}
   image:
-    registry: docker.io
-    repository: mendersoftware/deviceauth
+    registry: ""
+    repository: ""
     tag: mender-master
     imagePullPolicy: IfNotPresent
   service:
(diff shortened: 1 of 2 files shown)


mender/values.yaml
```diff
@@ -95,8 +95,8 @@ device_auth:
       memory: 128M
   affinity: {}
   image:
-    registry: docker.io
-    repository: mendersoftware/deviceauth
+    registry: ""
+    repository: ""
```
tranchitella, line 99: Why are you setting these variables to empty string?

MaciejTe: All enterprise services have empty strings in `registry` and `repository`. I fixed if statement in deviceauth deplyoment so proper values are chosen both when OS and enterprise template is prepared. Thanks!

tranchitella (approved):

PR #59, opened 2021-11-11
fixed a faulty flag to allow showing deviceconfig options in device details in the gui

mzedel:


mzedel:
@lluiscampos should that be backported to `3.1.x`, `3.0.x` and `2.7.x`?

tranchitella (approved):

tranchitella:
@mzedel yes, I can take care of that

PR #63, opened 2021-12-17
Replace api-gateway TCP probes with readiness and liveness probes

MaciejTe:
Replace api-gateway TCP probes with readiness and liveness probes

Changelog: None

Signed-off-by: Maciej Tomczuk <maciej.tomczuk@northern.tech>

### DIFF
--- a/mender/templates/api-gateway-configmap.yaml
+++ b/mender/templates/api-gateway-configmap.yaml
@@ -20,6 +20,26 @@ data:
       # Routers
       #
       routers:
+{{- if .Values.api_gateway.env.SSL }}
+        #
+        # redirection from http to https
+        #
+        redirect-to-https:
+          entrypoints: http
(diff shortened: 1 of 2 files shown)


mender/templates/api-gateway-configmap.yaml
```diff
@@ -20,6 +20,26 @@ data:
       # Routers
       #
       routers:
+{{- if .Values.api_gateway.env.SSL }}
+        #
+        # healthz
```
tranchitella, line 25:
```suggestion
        # redirection from http to https
```

tranchitella (changes requested):

mender/templates/api-gateway-configmap.yaml
MaciejTe: done!

tranchitella (approved):

PR #67, opened 2022-01-08
Fix: docker image name for tenantadm

tranchitella:
Changelog: none


tranchitella:
I am merging this fix without waiting for a review, as it is affects only the documentation and the change is straightforward.

PR #73, opened 2022-01-25
Fix readiness probe for api-gateway deployment

benjamin-texier:
The "periodseconds" field is present several times in the api-gateway readiness probe.

See [MEN-5407](https://tracker.mender.io/browse/MEN-5407)


tranchitella (approved):
Thanks for your contribution

PR #77, opened 2022-03-07
Allow "nodeSelector" option on all deployments

akos-denke-arm:
Allow to select the node via helm values.

This contribution allows to use this chart in Kubernetes clusters, where the default architecture is not x86_64 for the nodes.
With this change, the users will be able to select their own nodes other than the default one, which can be based on x86_64 as that is the only supported architecture as far I see.
The node selection is done based on a pattern widely used among popular helm charts.

Some examples from other helm charts:

ingress-nginx: https://github.com/kubernetes/ingress-nginx/search?q=nodeSelector
* https://github.com/kubernetes/ingress-nginx/blob/15b0aba03b700daacf0e9a3f5154ca1b9f77ee18/charts/ingress-nginx/templates/controller-deployment.yaml#L189
* https://github.com/kubernetes/ingress-nginx/blob/15b0aba03b700daacf0e9a3f5154ca1b9f77ee18/charts/ingress-nginx/templates/controller-daemonset.yaml#L191
* ...

minio: https://github.com/minio/charts/search?q=nodeSelector
* https://github.com/minio/charts/blob/a5c84bcbad884728bff5c9c23541f936d57a13b3/minio/templates/deployment.yaml#L184
* https://github.com/minio/charts/blob/a5c84bcbad884728bff5c9c23541f936d57a13b3/minio/templates/statefulset.yaml#L136
* ...

But you can see this pattern in almost all charts:
https://github.com/helm/charts/search?q=nodeSelector

I hope you will find this contribution useful!

### DIFF
--- a/README.md
+++ b/README.md
@@ -188,6 +188,7 @@ The following table lists the parameters for the `api-gateway` component and the
 | `api_gateway.image.repository` | Docker image repository | `traefik` |
 | `api_gateway.image.tag` | Docker image tag | `v2.5` |
 | `api_gateway.image.imagePullPolicy` | Docker image pull policy | `IfNotPresent` |
+| `api_gateway.nodeSelector` | [Node selection](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#nodeselector) | `{}` |
 | `api_gateway.replicas` | Number of replicas | `1` |
 | `api_gateway.affinity` | [Affinity map](https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#affinity-and-anti-affinity) for the POD | `{}` |
 | `api_gateway.resources.limits.cpu` | Resources CPU limit | `600m` |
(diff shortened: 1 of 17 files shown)


akos-denke-arm:
@tranchitella @kacf Can you please take a look on this small but useful PR?

tranchitella:
@akos-denke-arm thank you for your contribution, it looks good! Can I kindly ask you to update the README, too?

akos-denke-arm:
@tranchitella Sure, I will push the README update soon!

akos-denke-arm:
@tranchitella Here is the second round! I also added the new iot_manager microservice to the change.
Can you please elaborate, why the CI is failing so I can fix it? I got 404 on the GitLab project even if I am logged in. Probably, I do not have permission to see it.

tranchitella:
Thanks @akos-denke-arm, you need to sign off your second commit.

akos-denke-arm:
Thanks! You are totally right! Here we go!

PR #80, opened 2022-03-16
fix: Replace Hosted Mender references with `global.url` in workflows

alfrunes:
Depends on mendersoftware/workflows-enterprise#359

### DIFF
--- a/mender/templates/create-artifact-worker-deploy.yaml
+++ b/mender/templates/create-artifact-worker-deploy.yaml
@@ -57,6 +57,8 @@ spec:
         env:
         - name: WORKFLOWS_NATS_URI
           value: "{{ .Values.global.nats.URL }}"
+        - name: WORKFLOWS_MENDER_URL
+          value: "{{ .Values.global.url }}"
         - name: CREATE_ARTIFACT_GATEWAY_URL
           value: {{ .Values.global.url | default (ternary (printf "https://%s" .Values.api_gateway.service.name ) (printf "http://%s" .Values.api_gateway.service.name) (.Values.api_gateway.env.SSL)) }}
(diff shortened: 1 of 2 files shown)


alfrunes:
@merlin-northern this doesn't make the email templates fully customizable, but at least it updates the reference to the configured server url. Should we expose all workflows email templates as customizable variables?

tranchitella:
@alfrunes honestly, I don't think so.

tranchitella (approved):

merlin-northern:
> @merlin-northern this doesn't make the email templates fully customizable, but at least it updates the reference to the configured server url. Should we expose all workflows email templates as customizable variables?

we had no request for that. assuming you have your own template there is a similar amount of work required to provide it in the env or to replace it in the workflows pod, right? so I believe we are fine here. I have also forwarded you the original request, just as a refresher, we discussed that a long time ago.

merlin-northern (commented):
the change looks good, but see my comment.

merlin-northern (approved):

alfrunes:
> my question is, what does this change? the original problem: onprem installs send 2fa and password reset emails with hard coded server url, this is still the case:
> 
> ```shell
> # cat worker/password_reset_email.txt
> Dear Mender User,
> 
> You recently requested to reset your password for your Mender account.
> Click the link below or copy and paste the link into your browser
> to reset your password:
> 
> https://hosted.mender.io/ui/#/password/${workflow.input.secret_hash}
> ```
> 
> we need some followups in the workflows, right?

This is exactly what the accompanying PR (mendersoftware/workflows-enterprise#359) is doing, it replaces all hard-coded references to Hosted Mender with a template variable exposed as an environment variable.

PR #81, opened 2022-04-01
feat(rate limits): set default rate limits and caching in Mender Enterprise

tranchitella:
Add a redis pod and configure deviceauth and useradm to connect to it
enabling rate limits and caching of JWT tokens. At the same time, set
default rate limits in tenantadm.

Ticket: MEN-5270

### DIFF
--- a/mender/templates/api-gateway-configmap.yaml
+++ b/mender/templates/api-gateway-configmap.yaml
@@ -47,6 +47,7 @@ data:
         auditlogs:
           entrypoints: {{ $scheme }}
           middlewares:
+          - ratelimit
           - userauth
           - sec-headers
           - compression
@@ -62,6 +63,7 @@ data:
         iot-manager:
           entrypoints: {{ $scheme }}
           middlewares:
+          - ratelimit
           - userauth
           - sec-headers
           - compression
@@ -77,6 +79,7 @@ data:
         deployments:
           entrypoints: {{ $scheme }}
           middlewares:
+          - ratelimit
           - devauth
           - sec-headers
           - compression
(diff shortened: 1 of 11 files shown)


MaciejTe (approved):

mender/templates/api-gateway-configmap.yaml
```diff
@@ -473,6 +495,15 @@ data:
...
           compress: true
 
+        ratelimit:
+          ratelimit:
+            average: {{ .Values.api_gateway.ratelimit.average }}
+            burst: {{ .Values.api_gateway.ratelimit.burst }}
+            period: {{ .Values.api_gateway.ratelimit.period }}
+            sourcecriterion:
+              ipstrategy:
+                depth: 1
```
alfrunes, line 505: Doesn't this parameter depend on the infrastructure you're running Mender on-prem? Should we make it configurable?

alfrunes (approved):
Thank you for taking care of this :pray:
The PR looks good, I only have one nitpicky question.

mender/templates/api-gateway-configmap.yaml
```diff
@@ -473,6 +495,10 @@ data:
         compression:
           compress: true
 
+        ratelimit:
+          ratelimit:
```
alfrunes, line 499: I believe this is case sensitive:
```suggestion
          rateLimit:
```

tranchitella: Fixed, it was a copy and paste from `saas`.
https://github.com/mendersoftware/saas/pull/672

PR #87, opened 2022-05-16
feat: allow to use external secrets

benjamin-texier:
This merge request allows to use secrets for mender deployment.

This makes it easier to deploy Mender with the Postgresql database, Nats and Minio. It is then easier to deploy mender without having to manage passwords.

### DIFF
--- a/mender/templates/secret-s3-artifacts.yaml
+++ b/mender/templates/secret-s3-artifacts.yaml
@@ -1,3 +1,4 @@
+{{ if not .Values.global.s3.existingSecret }}
 apiVersion: v1
 kind: Secret
 metadata:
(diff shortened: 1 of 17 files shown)


mender/templates/secret-s3-artifacts.yaml
```diff
@@ -1,3 +1,4 @@
+{{ if not .Values.global.s3.existingSecret }}
```
oldgiova, line 1: The closing `{{ end }}` is missing

oldgiova (changes requested):
Hi @benjamin-texier , thanks for your awesome contribution!

Could you also please add the new `existingSecret` parameters in the README.md? Then we're good to test it

oldgiova:
Thanks @benjamin-texier for the contribution! I moved your commit to this new PR https://github.com/mendersoftware/mender-helm/pull/206 so I'm able to complete your and add small things

PR #90, opened 2022-06-09
Fix/improve the s3 values

alfrunes:


alfrunes:
@tranchitella I also added a commit mirroring the service ports to the Træfik entrypoints.

tranchitella (approved):

PR #92, opened 2022-06-23
feat: Configurable container ports for api-gateway

WalterMoar:
The commit "feat: Reuse service port for api-gateway entrypoints" in #90 made the `api_gateway` deployment configurable using the service ports. However, I ran into some problems using this:
1. In a Red Hat OpenShift cluster, the container ports cannot be privileged (<1024) ports, so it is common to have the service on privileged ports (80/443) and the container listening on non-privileged ports (8888/8443).
2. The two containerPort settings and readiness/liveness/startup probes are not configurable but need to use the same ports.

I added new settings `api_gateway.httpPort` and `api_gateway.httpsPort` that are used to configure all the ports in the deployment, and also the target ports in the service.

### DIFF
--- a/README.md
+++ b/README.md
@@ -193,6 +193,8 @@ The following table lists the parameters for the `api-gateway` component and the
 | `api_gateway.podAnnotations` | add custom pod annotations | `nil` |
 | `api_gateway.replicas` | Number of replicas | `1` |
 | `api_gateway.affinity` | [Affinity map](https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#affinity-and-anti-affinity) for the POD | `{}` |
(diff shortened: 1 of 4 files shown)


WalterMoar:
@tranchitella - The Gitlab checks failed for this, but I'm getting a 404 when I try to get the Details. Is it possible that these checks are not working?

tranchitella:
@WalterMoar I'm sorry to be pedantic, but our CI pipelines complains about the mismatch between the Author and the Sign-off. The author name in the commit doesn't have a space between your first and last names. Can you please make them to match?

WalterMoar:
@tranchitella Oh, that's why the check was failing? Fixed! Thank you.

PR #106, opened 2022-11-11
fix: Upgrade to Mender 3.3.1.

kacf:
Changelog: Title
Ticket: None

Signed-off-by: Kristian Amlie <kristian.amlie@northern.tech>


kacf:
A bit late, but finally catching up with the last items.

alfrunes (approved):

PR #116, opened 2023-02-09
refac: Reindex reporting hook Helm values and naming

alfrunes:


oldgiova (changes requested):
Adding `tolerations` as well

alfrunes:
Thanks for the suggestion @oldgiova, I amended my commit with the suggested changes.

oldgiova (approved):

PR #117, opened 2023-02-12
Add podSecurityContext and containerSecurityContext

philband:
This PR adds support for `securityContext` in both the pod and container spec. 

The newly added values each have an `enabled` switch, which is disabled by default for backward compatibility.

For the `api_gateway` securityContext to work, the internal listening ports have been moved out of the privileged port range.

The `gui` container can not work without upstream adjustments of the container image. Currently, the port used by nginx is fixed to `80`. I added `gui.httpPort` to be able to change the port when it gets configurable in the `gui` image.

Documentation for all added parameters is provided as well. 

Let me know if you require any adjustments.

### DIFF
--- a/README.md
+++ b/README.md
@@ -217,8 +217,8 @@ The following table lists the parameters for the `api-gateway` component and the
 | `api_gateway.podAnnotations` | add custom pod annotations | `nil` |
 | `api_gateway.replicas` | Number of replicas | `1` |
 | `api_gateway.affinity` | [Affinity map](https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#affinity-and-anti-affinity) for the POD | `{}` |
-| `api_gateway.httpPort` | Port for the HTTP listener in the container | `80` |
-| `api_gateway.httpsPort` | Port for the HTTPS listener in the container | `443` |
+| `api_gateway.httpPort` | Port for the HTTP listener in the container | `9080` |
+| `api_gateway.httpsPort` | Port for the HTTPS listener in the container | `9443` |
 | `api_gateway.resources.limits.cpu` | Resources CPU limit | `600m` |
 | `api_gateway.resources.limits.memory` | Resources memory limit | `1G` |
 | `api_gateway.resources.requests.cpu` | Resources CPU request | `600m` |
@@ -239,7 +239,12 @@ The following table lists the parameters for the `api-gateway` component and the
 | `api_gateway.rateLimit.burst` | See the [Traefik rate limit configuration options](https://doc.traefik.io/traefik/v2.6/middlewares/http/ratelimit/#configuration-options) | `100` |
 | `api_gateway.rateLimit.period` | See the [Traefik rate limit configuration options](https://doc.traefik.io/traefik/v2.6/middlewares/http/ratelimit/#configuration-options) | `1s` |
 | `api_gateway.rateLimit.sourceCriterion` | See the [Traefik rate limit configuration options](https://doc.traefik.io/traefik/v2.6/middlewares/http/ratelimit/#configuration-options) | `{"ipStrategy": {"depth": 1}}` |
-
+| `api_gateway.podSecurityContext.enabled` | Enable [security context](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/) | `false` |
+| `api_gateway.podSecurityContext.runAsNonRoot` | Run as non-root user | `true` |
+| `api_gateway.podSecurityContext.runAsUser` | User ID for the pod | `65534` |
+| `api_gateway.containerSecurityContext.enabled` | Enable container [security context](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/) | `false` |
+| `api_gateway.containerSecurityContext.allowPrivilegeEscalation` | Allow privilege escalation for container | `false` |
+| `api_gateway.containerSecurityContext.runAsUser` | User ID for the container | `65534` |
 
 ### Parameters: deployments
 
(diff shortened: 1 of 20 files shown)


mzedel:
Thanks for this @philband !
For the change to the GUI image there might be some context [here](https://hub.mender.io/t/problems-with-mender-helm-chart-on-openshift-privileged-ports/5041/3)...

philband:
Thanks for the hint @mzedel !
For testing I rewrote the `Dockerfile` and `nginx.conf` currently in the `gui` repo: https://github.com/philband/mender-gui/commit/6e47a6f40fc620fdba05394984f0fba520e9cec7

This fix builds a container that can successfully run with fully dropped privileges and as non-root. However, changing the `gui` HTTP port away from `80` seems a lot less trivial. I have not such intimate knowledge of the entire codebase as to know where changes would be necessary.

Please let me know if I should make a PR to the `gui` repo for changing the port.

mzedel:
To reduce the need for changes in the gui image, there is the [unprivileged nginx image](https://hub.docker.com/r/nginxinc/nginx-unprivileged) that should contain most adjustments already.
However as stated in the Mender Hub link, the gui image will align with the nginx image and thus it might be best to get the main nginx image to incorporate the unprivileged changes there. I will go through the gui codebase to make sure there is nothing preventing the gui code from running on a different port.

philband:
@mzedel friendly ping on this. I just rebased on master and added the `securityContext`  for dbmigration.

mender/values.yaml
```diff
@@ -66,8 +66,8 @@ api_gateway:
       memory: 512M
   affinity: {}
   nodeSelector: {}
-  httpPort: 80
-  httpsPort: 443
+  httpPort: 9080
+  httpsPort: 9443
```
mzedel, line 70: please set these back to their defaults

mzedel (commented):
I'll leave the rest to @oldgiova and @tranchitella, but I would think we can't change the traefik ports without breaking existing installations...

mender/templates/api-gateway-deploy.yaml
```diff
@@ -42,11 +42,17 @@ spec:
       {{- with .Values.api_gateway.affinity }}
       affinity: {{ tpl (toYaml .) $ | nindent 8 }}
       {{- end }}
+{{- if .Values.api_gateway.podSecurityContext.enabled }}
```
oldgiova, line 45: Suppose the case where the Security Context is enable but still the httpPort is 80 or 443 (privileged ports): the helm release won't work.

I'd suggest to set the following check on top of the `api-gateway-deploy.yaml` file: the helm deployment won't start with this setting.

```
{{- if or .Values.api_gateway.podSecurityContext.enabled .Values.api_gateway.containerSecurityContext.enabled }}
  {{- if lt (int .Values.api_gateway.httpPort) 1024  }}
    {{- fail ".Values.api_gateway.httpPort can't be less than 1024 when Security Context is enabled" }}
  {{- end }}
{{- end }}
```

mender/templates/gui-deploy.yaml
```diff
@@ -42,22 +42,28 @@ spec:
...
         # Readiness/liveness probes
         livenessProbe:
           tcpSocket:
-            port: 80
+            port: {{ .Values.gui.httpPort }}
           periodSeconds: 5
         readinessProbe:
           tcpSocket:
-            port: 80
+            port: {{ .Values.gui.httpPort }}
```
oldgiova, line 66: Unfortunately the `gui` container is not currently supporting non-root users. Please remove the modifications here for now.

mender/templates/gui-svc.yaml
```diff
@@ -31,7 +31,7 @@ spec:
   ports:
   - port: {{ .Values.gui.service.port }}
     protocol: TCP
-    targetPort: 80
+    targetPort: {{ .Values.gui.httpPort }}
```
oldgiova, line 34: Unfortunately the `gui` container is not currently supporting non-root users. Please remove the modifications here for now.

mender/values.yaml
oldgiova: After a deep check I can confirm this modification works properly and you can keep it.

README.md
```diff
@@ -337,6 +354,13 @@ The following table lists the parameters for the `gui` component and their defau
 | `gui.service.loadBalancerSourceRanges` | Service load balancer source ranges | `nil` |
 | `gui.service.port` | Port for the service | `80` |
 | `gui.service.nodePort` | Node port for the service | `nil` |
+| `gui.httpPort` | Port for the HTTP listener in the container | `80` |
+| `gui.podSecurityContext.enabled` | Enable [security context](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/) | `false` |
+| `gui.podSecurityContext.runAsNonRoot` | Run as non-root user | `true` |
+| `gui.podSecurityContext.runAsUser` | User ID for the pod | `65534` |
+| `gui.containerSecurityContext.enabled` | Enable container [security context](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/) | `false` |
+| `gui.containerSecurityContext.allowPrivilegeEscalation` | Allow privilege escalation for container | `false` |
+| `gui.containerSecurityContext.runAsUser` | User ID for the container | `65534` |
```
oldgiova, line 363: Unfortunately the `gui` service is not supporting non-root users. Please set back to default for now.

oldgiova (changes requested):
Thanks @philband for your valuable contribution! it's mostly good for me, but unfortunately we cannot use the Security Context feature with current `gui` service implementation, so I'm asking you to set back to defaults the `gui` modifications.
Can you also please add a failure check in the `api-gateway-deploy` as per my suggestion?

Thank you very much for your contribution!

README.md
philband: @oldgiova thank you for the detailed feedback. I would argue that we should give users the option to use a custom-built `gui` image which can run rootless like we do already. The change just adds options to configure the ports, but does not change the default. Also, the securityContext is disabled by default.

oldgiova: > @oldgiova thank you for the detailed feedback. I would argue that we should give users the option to use a custom-built `gui` image which can run rootless like we do already. The change just adds options to configure the ports, but does not change the default. Also, the securityContext is disabled by default.

:thinking: yeah, you're right. Let's keep it then.

mender/templates/api-gateway-deploy.yaml
philband: I added checks for both the http and https port as suggested.

oldgiova (approved):
Thank you again @philband for your kind contribution!

PR #122, opened 2023-03-09
chore: Update service versions to Mender 3.5.1 release.

kacf:
Changelog: Title
Ticket: None

### DIFF
--- a/mender/templates/gui-deploy.yaml
+++ b/mender/templates/gui-deploy.yaml
@@ -62,7 +62,7 @@ spec:
 
         env:
         - name: INTEGRATION_VERSION
-          value: "3.5.0"
+          value: "3.5.1"
         - name: MENDER_VERSION
           value: "3.5.0"
         - name: MENDER_ARTIFACT_VERSION
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ Using `helm`:
 
 ```bash
 $ make package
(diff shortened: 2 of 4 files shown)


mender/templates/gui-deploy.yaml
```diff
@@ -62,7 +62,7 @@ spec:
 
         env:
         - name: INTEGRATION_VERSION
-          value: "3.5.0"
+          value: "3.5.1"
         - name: MENDER_VERSION
           value: "3.5.0"
```
kacf, line 67: @mzedel: Can you confirm that this is correct? The Mender client version is still 3.5.0, IOW the Debian package. But the QEMU image is still part of the release, and is therefore tagged as `mender-3.5.1`. Will it treat it correctly?

kacf:
Whoever reviews this, can I ask that you also do a quick sanity check that it runs?

mender/templates/gui-deploy.yaml
mzedel: Looks correct to me!
The client version should only be relevant for the version information in the lower left corner, the device connection snippets all rely on what the install script installs (afaik the latest stable release) or use the untagged docker image (so `latest`)...

kacf: Ok, that should all work! Thanks!

tranchitella (approved):

PR #125, opened 2023-03-14
Update version references for Mender 3.3.2.

kacf:


kacf:
I would appreciate a sanity check that it works as well.

tranchitella (approved):

PR #126, opened 2023-03-14
Global toleration and affinity

oldgiova:
Ticket: QA-583

Added default toleration and affinity for `mender-helm-test`. With the `merge` strategy, values can be overridden by specific values.


mender/templates/api-gateway-deploy.yaml
```diff
@@ -1,4 +1,5 @@
 {{- if .Values.api_gateway.enabled }}
+{{- $merged := merge (deepCopy .Values.api_gateway) (deepCopy (default (dict) .Values.global)) -}}
```
alfrunes, line 2: Should we instead create a new `default` section? The `global` values are shared across sub-charts and already contain a lot of attributes that are not used at the service level.

alfrunes: I would prefer the first option to avoid potential name collisions.

oldgiova: Got it. Let me fix

oldgiova: Done, you can review it again

alfrunes (approved):

PR #129, opened 2023-04-06
chore: added quick kubeconform tests

oldgiova:


tranchitella (approved):
What about running them in the gitlab pipeline?

oldgiova:
> What about running them in the gitlab pipeline?

Yes, it's on the roadmap

PR #139, opened 2023-04-26
**WIP: don't merge** Revert 3.5 changes into master

oldgiova:
All the 3.5 features have been reverted


alfrunes:
I'm not sure if reverting all the features are necessary, or we should just leave out reporting/Opensearch?
Besides, should we wait for the reporting dependency issue in gui to be fixed? In the current state a deployment using master branches won't work.

tranchitella (approved):

tranchitella (changes requested):
Sorry, wrong button. This is not the right approach.
What we want to do is delivering master without reporting and opensearch.
This won't fully work until UI decouples from reporting, but we'll have it relatively soon.

oldgiova:
Ok, If I understood correctly then:
* We want to keep the `generate_delta_worker`  feature
* We have to remove the reporting service (and its reindexing jobs) and Opensearch
* We have to wait for the UI deployment.


If these are correct, then I'm going to prepare a new PR, branched from `3.5.x`, without reporting and opensearch.

PR #141, opened 2023-04-27
refac: Replace Traefik dynamic config provider from file to k8s CRDs

alfrunes:
- [x] TODO: This commit does not migrate the minio values

Remark:
 * The gui error respoinder had to be exposed under the fixed port 8080 to make this work (this is also the expected upstream port).
 * The error responders are removed from the devices API
   * These requests are not parsed by the frontend.
 * Tested with Traefik v10


mender/crds/ingressroutes.traefik.containo.us.yaml
oldgiova, line 1: Did you considered using a sub-chart? https://doc.traefik.io/traefik/getting-started/install-traefik/#use-the-helm-chart
This for ease on maintenance, mainly.

oldgiova (approved):
Really cool stuff! LGTM. I'm just a bit scared of maintainability of this solution because it adds complexity and another component. Is this addressing a known issue we had?

mzedel:
> ... Is this addressing a known issue we had?

If I'm not mistaken, this would make it easier to get feature branch deployments for e.g. GUI PRs - which is a long standing wish from reviewers to help reduce review times...

PR #150, opened 2023-05-05
Global registry.image.tag

oldgiova:
Using a `global.registry.image.tag` for Mender deployment (defult to `mender-3.4`) so to not specify it in every deployment.
This leads also a better user experience with selecting a different Mender version within the same Helm Chart.

Ticket: https://northerntech.atlassian.net/browse/MC-6798

### DIFF
--- a/README.md
+++ b/README.md
@@ -141,6 +141,7 @@ The following table lists the global parameters supported by the chart and their
 | `global.image.registry` | Global Docker image registry | `registry.mender.io` |
 | `global.image.username` | Global Docker image registry username | `nil` |
 | `global.image.password` | Global Docker image registry username | `password` |
+| `global.image.tag` | Global Docker image registry tag | `mender-3.4` |
 | `global.imagePullSecrets` | Global Docker registry secret names as an array | `[]` (does not add image pull secrets to deployed pods)  |
 | `global.mongodb.URL` | MongoDB URL | `mongodb://mongodb` |
 | `global.nats.URL` | NATS URL | `nats://nats:4222` |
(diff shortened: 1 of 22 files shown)


README.md
```diff
@@ -750,7 +751,7 @@ The following table lists the parameters for the `generate-delta-worker` compone
 | `generate_delta_worker.automigrate` | Enable automatic database migrations at service start up | `true` |
 | `generate_delta_worker.image.registry` | Docker image registry | `registry.mender.io` |
 | `generate_delta_worker.image.repository` | Docker image repository | `mendersoftware/generate-delta-worker` |
-| `generate_delta_worker.image.tag` | Docker image tag | `mender-3.5.0` |
+| `generate_delta_worker.image.tag` | Docker image tag | `nil` |
```
merlin-northern, line 754: there is also this: [tag: mender-3.5.0](https://github.com/mendersoftware/mender-helm/blob/master/mender/values.yaml#L200)

README.md
```diff
@@ -712,7 +713,7 @@ The following table lists the parameters for the `devicemonitor` component and t
 | `devicemonitor.automigrate` | Enable automatic database migrations at service start up | `true` |
 | `devicemonitor.image.registry` | Docker image registry | `registry.mender.io` |
 | `devicemonitor.image.repository` | Docker image repository | `mendersoftware/devicemonitor` |
-| `devicemonitor.image.tag` | Docker image tag | `mender-3.5.0` |
+| `devicemonitor.image.tag` | Docker image tag | `nil` |
```
merlin-northern, line 716: I am wondering why do we default to null? should there be some default? arent we defaulting to global tag?

README.md
oldgiova: > there is also this: [tag: mender-3.5.0](https://github.com/mendersoftware/mender-helm/blob/master/mender/values.yaml#L200)

This is from the `generate_delta_worker` image, not available in `mender-3.4.0` and let disabled in the `values.yaml` by default. So we're overriding the default here.
I know it's not the best solution we have, but it's kind of "feature flag"

README.md
oldgiova: We could actually set mender tags by default, but would complex the setup a bit.
My proposal is just setting `mender-3.*.*` once in the global; for example: https://github.com/mendersoftware/mender-docs/pull/2102/files#diff-30ea1193d703087fda20ccd8993a3ae9bedca62db6d539e8833a7ba589026f65R62

merlin-northern (approved):

PR #153, opened 2023-05-18
:warning:  - PoC only don't merge: bump yaml files from upstream

oldgiova:
An upstream service pipeline triggers a `helm-version-bump` step: it gets required variables and update the `Chart.yaml` and `values.yaml` files for its `staging|MC-6387-helm-staging-poc` branch.

Ticket: MC-6387


oldgiova:
PoC is concluded.

PR #154, opened 2023-05-18
:warning:  - PoC only don't merge: sample staging branch

oldgiova:
This is the sample `staging` branch to be used to deploy to HM staging.
* For the sake of the PoC, it contains all the backing services as sub-charts (MongoDB, NATS, cert-manager)
* It contains the new pipeline step `helm-version-bump:staging-component` that update the `Chart.yaml` and `values.yaml` for the `staging|MC-6387-helm-staging-poc` branch

The update to HM process is decoupled: after updating the `Chart.yaml` and `values.yaml`, its job is finished. Later on, a Flux/Argocd process will pull updates from this upstream git branch and applies the new updates.


oldgiova:
PoC is concluded

PR #156, opened 2023-05-26
Sub-charts dependencies

oldgiova:
Proposal: use an external providers with sub-chart integration:
* redis
* cert-manager
* NATS
* MongoDB

Still todo: build the integration tests for the upgrade from 4.0.4 to 5.0.0 version - waiting for https://github.com/mendersoftware/mender-helm/pull/157


alfrunes (changes requested):
Strictly speaking, I think this change deserves a major version bump as it requires users to potentially change their helm values. The Redis parameters are moved into the global scope and no longer enabled by default, this should also be documented in the release notes/changelog.

oldgiova:
Superseded by: https://github.com/mendersoftware/mender-helm/pull/161

PR #162, opened 2023-06-19
Conditionally redirect middleware

oldgiova:
Re-applied the commit `ad123cc9c926bd52842fa00217c0232cefa10498` to be compliant to the current `master` branch, but also applied some conditional logic to let it working with older Mender releases as well.

### DIFF
--- a/mender/templates/_helpers.tpl
+++ b/mender/templates/_helpers.tpl
@@ -1,0 +1,29 @@
+{{/*
+Define Mender Major and minor version
+to be able to apply some conditional logic
+*/}}
+{{- define "menderVersionMinor" }}
+{{- with .Values.global.image }}
+{{- if contains "-" .tag }}
+  {{- $mndr_splitted := split "-" .tag -}}
+  {{- $mndr_version := split "." $mndr_splitted._1 }}
+  {{- printf "%s" $mndr_version._1 }}
+{{- else }}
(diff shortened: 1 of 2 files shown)


mender/templates/_helpers.tpl
```diff
@@ -0,0 +1,29 @@
...
+{{- if contains "-" .tag }}
+  {{- $mndr_splitted := split "-" .tag -}}
+  {{- $mndr_version := split "." $mndr_splitted._1 }}
+  {{- printf "%s" $mndr_version._0 }}
+{{- else }}
+  {{- $mndr_version := split "." .tag }}
+  {{- printf "%s" $mndr_version._0 }}
+{{- end }}
+{{- end }}
+{{- end }}
```
alfrunes, line 29: This is going to break once we start with rolling backend releases. I would simply update the chart version and break compatibility backward compatibility.

oldgiova: The rolling releases will update the services' tag, e.g.:
```
deployments:
  image:
    tag: master_1234566
```
And the `global.image.tag` could be a reference. We can even use the `appVersion` value for a general reference, but cannot be parametrized.

Anyway, I agree that this is not going to work in the long term with the rolling release... too messy

oldgiova:
This is not going to work with the upcoming rolling release method.

PR #163, opened 2023-07-01
Fix issues with mender setup when mongodb is replicaset

oldgiova:
Workaround for a known issue: https://github.com/bitnami/charts/issues/10843
Also documentation updated: https://github.com/mendersoftware/mender-docs/pull/2155

### DIFF
--- a/mender/values.yaml
+++ b/mender/values.yaml
@@ -62,11 +62,12 @@ mongodb:
   image:
     tag: "4.4.13-debian-10-r63"
   auth:
-    enabled: false
-    username: ""
-    password: ""
+    enabled: true
+    rootPassword: "pleasechangeme"
+    replicaSetKey: "pleasechangeme"
+    username: "mender"
     # reference to Bitnami chart only: the database name is managed by Mender only
-    database: ""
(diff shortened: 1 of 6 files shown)


mender/values.yaml
```diff
@@ -62,11 +62,12 @@ mongodb:
...
-    enabled: false
-    username: ""
-    password: ""
+    enabled: true
+    rootPassword: "pleasechangeme"
+    replicaSetKey: "pleasechangeme"
+    username: "mender"
     # reference to Bitnami chart only: the database name is managed by Mender only
-    database: ""
+    database: "mender"
```
alfrunes, line 70: This username/database does not make sense as I believe these are restricted to a single `database`. In our case, where we share credentials across all microservices, we only need in the `rootPassword` and `replicaSetKey`.

oldgiova: That was my first iteration, but then the bitnami chart complained for missing `database` object :shrug: .
That's why I set the comment: `# reference to Bitnami chart only: the database name is managed by Mender only`

alfrunes (approved):

PR #167, opened 2023-07-19
Revert "chore: added 3.6 upgrade tests"

oldgiova:
This reverts commit dd0e3510bcf518ea127c6d46b9eeff7a20174f8c.

Too early merge: it has to be merged with the 3.6 branch


merlin-northern (commented):
commit checker should not fail for the default git commit messages, I remember this discussion. how are we doing the reverts so it passes?

merlin-northern (approved):

oldgiova:
> commit checker should not fail for the default git commit messages, I remember this discussion. how are we doing the reverts so it passes?

I just amended it and force pushed :man_shrugging:

PR #168, opened 2023-07-20
feat: added internal Ingress resource

oldgiova:
Passing test: https://gitlab.com/Northern.tech/Mender/mender-helm/-/jobs/4717841092#L331

Depends on: https://github.com/mendersoftware/mender-test-containers/pull/382


mzedel:
@oldgiova doesn't this duplicate what's in https://github.com/mendersoftware/mender-helm/pull/141 ?

oldgiova:
> @oldgiova doesn't this duplicate what's in #141 ?

As per my understanding that PR is replacing the api configmap, whereas this one is adding the L7 ingress resource, so basically replacing [this manual installation step](https://docs.mender.io/3.4/server-installation/production-installation-with-kubernetes/mender-server#exposing-the-service).

Am I wrong?

tranchitella (approved):

PR #172, opened 2023-07-21
chore: aligned service versions with integration

mzedel:
to align with
https://github.com/mendersoftware/integration/blob/master/docker-compose.yml#L60
https://github.com/mendersoftware/integration/blob/master/docker-compose.yml#L270

### DIFF
--- a/mender/values.yaml
+++ b/mender/values.yaml
@@ -103,7 +103,7 @@ api_gateway:
   image:
     registry: docker.io
     repository: traefik
-    tag: v2.5
+    tag: v2.8
     imagePullPolicy: IfNotPresent
   replicas: 1
   resources:


oldgiova:
LGTM, let's also ask @tranchitella because he's the release manager

oldgiova (approved):

tranchitella:
@mzedel and @oldgiova, we are running 4.4.x in production now.
I prefer to release Mender 3.6 (LTS) with this version and leave the upgrade to 5.0.x in the next release cycle.
What do you think?

mender/values.yaml
```diff
@@ -103,7 +103,7 @@ api_gateway:
   image:
     registry: docker.io
     repository: traefik
-    tag: v2.5
+    tag: v2.8
```
tranchitella, line 106: This is okay.

mender/values.yaml
```diff
@@ -60,7 +60,7 @@ mongodb:
     service:
       type: LoadBalancer
   image:
-    tag: "4.4.13-debian-10-r63"
+    tag: "5.0"
```
tranchitella, line 63: This one I don't want to include in the release.

tranchitella (changes requested):

tranchitella (approved):

tranchitella:
Thanks @mzedel, I merged.

PR #182, opened 2023-07-31
feat: Upgrade MongoDB dependency to 5.0

alfrunes:


mender/values.yaml
```diff
@@ -88,7 +88,7 @@ mongodb:
     service:
       type: LoadBalancer
   image:
-    tag: "4.4.13-debian-10-r63"
+    tag: "5.0.19-debian-11-r13"
```
merlin-northern, line 91: as a review I checked that this tag exists.
only one qiestion, why are using this one, and not plain [5.0.19](https://hub.docker.com/layers/bitnami/mongodb/5.0.19/images/sha256-b76bbf58377eef8f71773b4c19e51405ae6fbd32b8cca8850877e5d810776d45?context=explore)? (no strings attached :) just wanted to know)

merlin-northern (approved):
lgtm. left one comment/question.

tranchitella (approved):

mender/values.yaml
alfrunes: If you have `imagePullPolicy: IfNotPresent` the node will not pull a new node image if the Debian version for that image changes.

PR #189, opened 2023-08-20
Added Port for the Error server block on the gui

oldgiova:
To align HM with the Helm Chart: this value is already present on the HM instance

Ticket: MC-6856

### DIFF
--- a/mender/values.yaml
+++ b/mender/values.yaml
@@ -318,6 +318,7 @@ gui:
     annotations: {}
     type: ClusterIP
     port: 80
+    portError: 8080
   httpPort: 80
   podSecurityContext:
     enabled: false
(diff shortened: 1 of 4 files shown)


mender/values.yaml
```diff
@@ -316,6 +316,7 @@ gui:
     annotations: {}
     type: ClusterIP
     port: 80
+    portError: ""
```
mzedel, line 319: should this be `8080` by default?

oldgiova: As of now it's disabled on the Helm chart, so by setting it it will be enabled starting from the next release. 
I guess it's fine and don't break anything, tough, so it's safe to set it

mzedel: ah sorry, only now I realize that this was for internal usage by the error middleware only (and traefik would return these errors through the regular port already), so is not really meant to be used by the outside world... hmm then I'm not sure if a different setting will have an effect there, but aligning it with the HM instance should be good in any case 👍...

mzedel (approved):

PR #191, opened 2023-08-23
Align HM/Helm - Deployments and PDB

oldgiova:
Focusing on the `kind: Deployments` drift. Here's the changes:
* Added support to external Image Pull Secrets, global and per-service
* Added support to extraArgs to the `api_gateway` service
* Traefik updated to `v2.9.6`
* You can now add pre-existing `priorityClassName` to the resources
* Added PodDisruptionBudget resources

### DIFF
--- a/README.md
+++ b/README.md
@@ -191,6 +191,8 @@ The following table lists the global, default, and other parameters supported by
 | `default.hpa.maxReplicas` | HorizontalPodAutoscaler maxReplicas | `nil` |
 | `default.hpa.metrics` | HorizontalPodAutoscaler metrics as defined in the [reference](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.28/#metricspec-v2-autoscaling) | `nil` |
 | `default.hpa.behavior` | HorizontalPodAutoscaler behavior as defined in the [reference](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.28/#horizontalpodautoscalerbehavior-v2-autoscaling) | `nil` |
+| `default.pdb.enabled` | PodDistruptionBudget enabled | `false` |
+| `default.pdb.minAvailable` | PodDistruptionBudget minAvailable | `1` |
 | `default.imagePullSecrets` | Optional list of existing Image Pull Secrets in the format of `- name: my-custom-secret` | `[]` |
 | `ingress.enabled` | Optional Mender Ingress | `false` |
 | `dbmigration.enable` | Helm Chart hook that trigger a DB Migration utility just before an Helm Chart install or upgrade  | `true` |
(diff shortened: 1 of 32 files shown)


README.md
```diff
@@ -150,11 +150,12 @@ The following table lists the global, default, and other parameters supported by
 | -------------------------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------- | -------------------------------------------------------- |
 | `global.enterprise` | Enable the enterprise features | `true` |
 | `global.hosted` | Enabled Hosted Mender specific features | `false` |
+| `global.priorityClassName` | Optional pre-existing priorityClassName to be assigned to the workwloads | `nil` |
 | `global.image.registry` | Global Docker image registry | `registry.mender.io` |
 | `global.image.username` | Global Docker image registry username | `nil` |
 | `global.image.password` | Global Docker image registry username | `password` |
 | `global.image.tag` | Global Docker image registry tag | `mender-3.6.2` |
-| `global.imagePullSecrets` | Global Docker registry secret names as an array | `[]` (does not add image pull secrets to deployed pods)  |
```
oldgiova, line 157: `global.imagePullSecrets` is currently not in use

mender/templates/_helpers.tpl
```diff
@@ -164,3 +164,24 @@ spec:
...
+kind: PodDisruptionBudget
+metadata:
+  name: {{ .name }}
+spec:
+  minAvailable: {{ $pdb.minAvailable | default 1 }}
+  selector:
+    matchLabels:
+      run: {{ .name }}
+{{- end }}
+{{- end }}
```
alfrunes, line 187: Nice :+1:

README.md
```diff
@@ -154,7 +154,7 @@ The following table lists the global, default, and other parameters supported by
...
+| `global.image.externalImagePullSecrets` | Optional list of existing Image Pull Secrets in the format of `- name: my-custom-secret` | `[]` |
```
alfrunes, line 158: I actually prefer the old (unused) name :see_no_evil: 
The `external` part seems redundant and "image" is repeated in the value key.

alfrunes (commented):
The Traefik version should be bumped to the latest `v2.10.4`.

Apart from that, feel free to ignore my naming preferences :see_no_evil:

README.md
oldgiova: > I actually prefer the old (unused) name 🙈 The `external` part seems redundant and "image" is repeated in the value key.

The thing is that `imagePullSecrets` is already taken:
* `global.imagePullSecrets`: unused
* `imagePullSecret`: its a template defined [here](https://github.com/mendersoftware/mender-helm/blob/master/mender/templates/_image-pull-secret.yaml)

What I want to do is to create this new ImagePullSecret, same as for HM, and don't interfere with the current one, created in with the `imagePullSecret` template.
Maybe something like `global.image.additionalPullSecrets` ?

oldgiova: I just found that `global.imagePullSecrets` is also inherited to subcharts. E.g. redis:
```
      imagePullSecrets:
        - name: map[name:docker-hub-secret-rollnrenew-robtest]
```
So I'm gonna move it to the `default` instead.

README.md
oldgiova: Correction: is used by subcharts (e.g.: Redis). I'd leave this out entirely to avoid mistakes

mender/templates/_image-pull-secret.yaml
```diff
@@ -1,3 +1,3 @@
-{{- define "imagePullSecret" }}
+{{- define "mender.enterprisePullSecret" }}
```
oldgiova, line 1: I chose this way: this template is used to define the imagePullSecrets for enterprise images only, so I just renamed it, to avoid confusions.

PR #192, opened 2023-08-25
feat: set Mender 3.6.2 image tags

tranchitella:


mender/Chart.yaml
```diff
@@ -1,8 +1,8 @@
 apiVersion: v2
-appVersion: "3.6.1"
+appVersion: "3.6.2"
 description: Mender is a robust and secure way to update all your software and deploy your IoT devices at scale with support for customization
 name: mender
-version: 5.2.2
+version: 5.2.3
```
oldgiova, line 5:
```suggestion
version: 5.2.4
```
For the Changelog, the version is `5.2.4`. It's still fine to me to use the `5.2.3` instead, since is not published yet.

tranchitella: Oh, sorry... I didn't notice it was unreleased.

oldgiova (approved):

PR #193, opened 2023-08-28
Fix/some issues

mzedel:
not yet sure if all of this is 100% right...

### DIFF
--- a/mender/values.yaml
+++ b/mender/values.yaml
@@ -1,5 +1,5 @@
 global:
-  enterprise: true
+  enterprise: false
   hosted: false
   auditlogs: true
   image:
--- a/mender/values.yaml
+++ b/mender/values.yaml
@@ -203,7 +203,7 @@ api_gateway:
 deployments:
   enabled: true
   podAnnotations: {}
-  automigrate: true
+  automigrate: false
   replicas: 1
   resources:
     limits:
@@ -244,7 +244,7 @@ deployments:
 device_auth:
   enabled: true
   podAnnotations: {}
-  automigrate: true
+  automigrate: false
   replicas: 1
   resources:
     limits:
@@ -290,7 +290,7 @@ device_auth:
 generate_delta_worker:
   enabled: false
   podAnnotations: {}
-  automigrate: true
+  automigrate: false
   replicas: 1
   resources:
     limits:
@@ -343,7 +343,7 @@ gui:
 inventory:
   enabled: true
   podAnnotations: {}
-  automigrate: true
+  automigrate: false
   replicas: 1
   resources:
     limits:
@@ -416,7 +416,7 @@ tenantadm:
 useradm:
   enabled: true
   podAnnotations: {}
-  automigrate: true
+  automigrate: false
   replicas: 1
   resources:
     limits:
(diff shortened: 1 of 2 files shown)


mender/values.yaml
```diff
@@ -10,7 +10,7 @@ global:
   mongodb:
     URL: mongodb://mongodb
   nats:
-    URL: nats://nats:4222
+    URL: nats://mender-nats:4222
```
oldgiova, line 13: When you set `nats.URL` you're using an external NATS, for e.g. one you deployed manually with a dedicated helm chart. Historically this was named just `nats`, so we kept the same name for backward compatibility: 
https://docs.mender.io/3.2/server-installation/production-installation-with-kubernetes/nats

by updating it, some customer setup could brake. 

If you deploy Mender with the integrated NATS (`nats.enabled: true`) you then have to let `nats.URL: ""`

oldgiova:
Just out of curiosity: why we want to switch to automigrate off? I mean, it makes sense to let just the migration job to do the job, but did you experienced a specific issue to propose the switch?

mender/values.yaml
oldgiova: In your own override values files, yes

mzedel: so does the repo not aim to allow a working installation with what's in the repo?
or rather: we provide the included nats chart, but if you use it values or example values file in the repo it won't start - is this intended?

mzedel:
> Just out of curiosity: why we want to switch to automigrate off? I mean, it makes sense to let just the migration job to do the job, but did you experienced a specific issue to propose the switch?

yep, the services didn't start (they still won't start since the migration job won't get scheduled due to resource limits - but that's a different issue)...

also the default for the services is not to have automigrate enabled

mender/values.yaml
oldgiova: Unfortunately yes: historically this helm chart never worked alone, but needed external backing services (NATS, Mongo...). We documented the NATS standalone helm chart setup in the docs (like [this](https://docs.mender.io/3.2/server-installation/production-installation-with-kubernetes/nats)).

When we added the integrated NATS, we could not break the default external NATS and Mongo: customers would have experienced issues with the upgrade.
So the best choice we took was to keep backward compatibility and let the default external NATS.

oldgiova:
> > Just out of curiosity: why we want to switch to automigrate off? I mean, it makes sense to let just the migration job to do the job, but did you experienced a specific issue to propose the switch?
> 
> yep, the services didn't start (they still won't start since the migration job won't get scheduled due to resource limits - but that's a different issue)...
> 
> also the default for the services is not to have automigrate enabled

I'm in favor to set automigrate disable, this also prevents other issues we had with customers (e.g. SLB). But since I could miss something about the backend, I'd like @alfrunes opinion here.

alfrunes:
Sorry, this one slipped through my cracks.
The automigrate options should be off by default since the migration job renders this a no-op, and this just adds additional startup time for new pods. Also, the whitespace truncation needs to be fixed anyway.

alfrunes (approved):

alfrunes (commented):
Although I'm all for these changes, the changing of the default value of  `global.enterprise` and `nats.URL` probably requires bumping the major chart version since existing installations need to explicitly overwrite this value. This also needs to be clear in the `Changelog.md`.

PR #210, opened 2023-10-12
fix: indent cronjob annotations correctly

vphoikka:
Hello!

I noticed that the helm chart cannot render the yaml correctly if you add pod annotations to the deployments section. I tracked this down to be a simple indentation miss.


oldgiova (approved):
Hello @vphoikka thanks for your contribution! LGTM!

PR #216, opened 2023-10-16
Split db migration containers

oldgiova:
Passing pipeline: https://gitlab.com/Northern.tech/Mender/mender-helm/-/pipelines/1039924156
second review round - passing pipeline: https://gitlab.com/Northern.tech/Mender/mender-helm/-/pipelines/1040729652

PR #217, opened 2023-10-31
feat: Add serviceAccount options to Helm Chart

bdomars:
Adds an option to use a named service account or create a named service account to use for all mender services.

Tried keeping this similar to how other Charts do it.

This is a nice to have if you use kubernetes service accounts to identify the pods in services like Vault for secret injection.

### DIFF
--- a/mender/templates/_helpers.tpl
+++ b/mender/templates/_helpers.tpl
@@ -185,3 +185,15 @@ spec:
       run: {{ .name }}
 {{- end }}
 {{- end }}
+
+{{/*
+Create the name of the service account
+*/}}
+{{- define "mender.serviceAccountName" -}}
+{{- if .Values.serviceAccount.create }}
+{{- default (include "mender.fullname" .) .Values.serviceAccount.name }}
+{{- else }}
+{{- default "default" .Values.serviceAccount.name }}
+{{- end }}
+{{- end }}
+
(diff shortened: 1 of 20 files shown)


oldgiova (commented):
Hi @bdomars, thanks for your contribution!

Unfortunately we just managed to merge a refactor of the deployments definition by including a `_podtemplate` that could be used for both deployments and migration jobs:
https://github.com/mendersoftware/mender-helm/pull/216

So you should rebase your changes and solve the conflicts. Sorry about that!

Please keep in mind that if you want to add the `serviceAccountName` to the `_podtemplate.yaml`, the service account must exists in the Helm hook pre-install, pre-upgrade phases:

```
    "helm.sh/hook": pre-install,pre-upgrade
    "helm.sh/hook-weight": "10"
    "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded
```
Otherwise the migration jobs won't run.

Suggestion: you could create two serviceaccounts: the ordinary one, and another one with the `-prerelease` suffix with helm hook annotations, then reference the two cases in the  `_podtemplate.yaml`

oldgiova:
I took your commit and solved the conflicts, and merged: https://github.com/mendersoftware/mender-helm/pull/231. This could be closed as superseded.
Thanks for you awesome contribution!

PR #223, opened 2023-12-12
Upgrade to Mender 3.7.0

tranchitella:


oldgiova (approved):
LGTM; also tested with the current `mender-master` tag

oldgiova:
Since you cannot use this chart to install the old 3.6.x, here's the updated doc for installing the old version: https://github.com/mendersoftware/mender-docs/pull/2288

PR #234, opened 2024-01-28
Switch to storage gateway from minio and CustomEnvs

oldgiova:
* Replaced minio specific gateway with a generic storage_gateway service, that could work for both minio and Amazon S3
* Added customEnvs options

Ticket: MC-6856

### DIFF
--- a/mender/CHANGELOG.md
+++ b/mender/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## master
 * Fix mongodb uri creation when using the mongodb subchart and replicast architecture
+* Added customEnv option to set default or per-service custom env variables
 
 ## Version 5.4.1
 * Upgrade to Mender version `3.7.1`.
--- a/mender/CHANGELOG.md
+++ b/mender/CHANGELOG.md
@@ -1,8 +1,20 @@
 # Mender Helm chart
 
-## master
+## 5.5.0
 * Fix mongodb uri creation when using the mongodb subchart and replicast architecture
 * Added customEnv option to set default or per-service custom env variables
+* Added generic `storage_proxy` service, that could
+  work for both minio and Amazon S3, and it's going to replace the `api_gateway.minio` configuration.
+* **Deprecations**:
+  * `api_gateway.minio` is deprecated in favor of `api_gateway.storage_proxy`.
+    This entry could be used, but it is no longer maintained, and could be removed
+    in future releases.  
+    **How to upgrade**: 
+    * set `api_gateway.minio.enabled=false`
+    * set `api_gateway.storage_proxy.enabled=true`
+    * set `api_gateway.storage_proxy.url` to the external storage url that you want to map externally. For example `https://fleetstorage.example.com`.
+      If you leave it empty, it uses the Amazon S3 external URL.
+
 
 ## Version 5.4.1
 * Upgrade to Mender version `3.7.1`.
--- a/mender/values.yaml
+++ b/mender/values.yaml
@@ -69,6 +69,11 @@ default:
     enabled: false
     minAvailable: 1
 
+  # custom default envs for every service
+  customEnvs: []
+  # - name: LOG_LEVEL
+  #   value: WARN 
+
 serviceAccount:
   create: false
   name:
@@ -290,6 +295,11 @@ deployments:
   #     cpu: 150m
   #     memory: 64M
 
+  # custom envs
+  customEnvs: []
+  # - name: LOG_LEVEL
+  #   value: DEBUG 
+
 device_auth:
   enabled: true
   podAnnotations: {}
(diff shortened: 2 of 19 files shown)


mender/templates/api-gateway/deployment.yaml
```diff
@@ -94,6 +94,12 @@ spec:
         resources:
 {{ toYaml .Values.api_gateway.resources | indent 10 }}
 
+        env:
+{{- if .Values.api_gateway.storage_proxy.enabled }}
+          - name: MENDER_STORAGE_PROXY
+            value: {{ include "mender.storageProxyUrl" . }}
+{{- end }}
```
alfrunes, line 101: What is this environment variable for?

mender/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## master
 * Fix mongodb uri creation when using the mongodb subchart and replicast architecture
+* Replaced minio specific gateway with a generic `storage_gateway` service, that could
```
alfrunes, line 5: This is a breaking change -- should deserve a major version bump in the `Chart.yaml`

mender/values.yaml
```diff
@@ -69,6 +69,11 @@ default:
     enabled: false
     minAvailable: 1
 
+  # custom default envs for every service
+  customEnvs: []
```
alfrunes, line 73: Can we `s/customEnvs/env/` to align it with Alvaldi?

mender/CHANGELOG.md
oldgiova: Right! Why not to use a drop-in replacement?

Keep using both the keys:
```
api_gateway:
  minio:
    enabled: true
```
and:
```
api_gateway:
  storage_proxy:
    enabled: false
```

mender/values.yaml
oldgiova: I tried, but I found two issues:
* The Alvaldi [utility](https://github.com/NorthernTechHQ/alvaldi-helm/blob/main/alvaldi/templates/_helpers.tpl#L261) is a bit different: it's not merging the envs (afaik) and I need a merge
* many keys `<service>.env` are already in place, with different notations:
  ```
    env:
      TENANTADM_MIDDLEWARE: prod
      TENANTADM_SERVER_PRIV_KEY_PATH: /etc/tenantadm/rsa/private.pem
  ```
But the Alvaldi notation implies:
```
env:
  - name: TENANTADM_MIDDLEWARE
    value: prod
  - name: TENANTADM_SERVER_PRIV_KEY_PATH
    value: /etc/tenantadm/rsa/private.pem
```

and this requires a delicate migration and a breaking change. Imho we should keep it simple with a key different than `env`, but let me know what you prefer.
If you want to align with Alvaldi, I'd prefer moving Alvaldi to `s/env/customEnvs/` or similar

alfrunes: Ahh, ok. I was not aware.

mender/templates/api-gateway/deployment.yaml
alfrunes: Indeed there it's used in the Traefik [dynamic configuration[(https://github.com/mendersoftware/saas/blob/master/kubernetes/mender_stack/gateway/config-mender-api-gateway-traefik.yaml#L7), but this environment variable does not affect the config here. With helm we can template this extra router directly although I would prefer having this at the ingress level and not inside Traefik.
IOW, I think we can drop this one and rely on the ingress.

oldgiova: Dropped, thanks

mender/CHANGELOG.md
```diff
@@ -1,7 +1,19 @@
...
+* Added generic `storage_proxy` service, that could
+  work for both minio and Amazon S3, and it's going to replace the `api_gateway.minio` configuration.
+* **Deprecations**:
+  * `api_gateway.minio` is deprecated in favor of `api_gateway.storage_proxy`.
+    This entry could be used, but it is no longer maintained, and could be removed
+    in future releases.  
+    **How to upgrade**: 
+    * set `api_gateway.minio.enabled=false`
+    * set `api_gateway.storage_proxy.enabled=true`
+    * set `api_gateway.storage_proxy.url` to the external storage url that you want to map externally. For example `https://fleetstorage.example.com`.
```
oldgiova, line 15: I'm not 100% sure about this statement

mender/CHANGELOG.md
```diff
@@ -1,7 +1,19 @@
 # Mender Helm chart
 
-## master
+## 5.5.0
 * Fix mongodb uri creation when using the mongodb subchart and replicast architecture
+* Added customEnv option to set default or per-service custom env variables
+* Added generic `storage_proxy` service, that could
+  work for both minio and Amazon S3, and it's going to replace the `api_gateway.minio` configuration.
+* **Deprecations**:
```
oldgiova, line 8: I chose a soft deprecation: `api_gateway.minio` is still enabled (and default choice) but it's considered deprecated.

alfrunes (approved):

PR #258, opened 2024-02-27
MC-7225: Upgrade MongoDB major version to 6.0

alfrunes:


alfrunes:
I'm not sure whether I should bump the chart major version or minor version is ok. This chart version provides a potential breaking change if upgrading from < 5.4 which is using MongoDB 4.4 since the cluster must be upgraded one major version at the time. On the other hand, this is just changing the default tag in the values files of an optional sub-chart.

tranchitella (approved):
I think 5.6.0 is okay.

oldgiova (approved):

oldgiova:
> I'm not sure whether I should bump the chart major version or minor version is ok. This chart version provides a potential breaking change if upgrading from < 5.4 which is using MongoDB 4.4 since the cluster must be upgraded one major version at the time. On the other hand, this is just changing the default tag in the values files of an optional sub-chart.

I think it's fine: the subchart is not recommended for production (as is written in the docs), so a minor version is ok.

PR #269, opened 2024-05-13
Fixes and alignements to saas

oldgiova:
* `generate_delta_worker`: don't enforce tags for the image.
* Added `api_gateway.accesslogs` parameter to enable/disable access logs.
* Bump traefik image to v2.11.2
* Move from megabytes to mebibytes for consistency.
* Added `inventory.mongodbExistingSecret` to override the default MongoDB secret.

### DIFF
--- a/mender/Chart.yaml
+++ b/mender/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: "3.7.4"
 description: Mender is a robust and secure way to update all your software and deploy your IoT devices at scale with support for customization
 name: mender
-version: 5.6.2
+version: 5.6.3
 keywords:
   - mender
   - iot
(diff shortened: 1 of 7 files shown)


mender/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: "3.7.4"
 description: Mender is a robust and secure way to update all your software and deploy your IoT devices at scale with support for customization
 name: mender
-version: 5.6.2
+version: 5.6.3
```
alfrunes, line 5: I would argue that since you're adding features to the helm chart, this should be a minor bump.
```suggestion
version: 5.7.0
```

alfrunes (approved):

mender/Chart.yaml
oldgiova: Right!

PR #270, opened 2024-05-14
chore: align mender with saas

oldgiova:
Ticket: MC-7302


alfrunes (approved):

alfrunes:
Would be great if you can review the tags from production/kustomize. I see the useradm-enterprise is outdated (there are a couple of follow-up releases made since Apr. 29th).

oldgiova:
> Would be great if you can review the tags from production/kustomize. I see the useradm-enterprise is outdated (there are a couple of follow-up releases made since Apr. 29th).

Done. I ran the missing deployment pipelines:  https://gitlab.com/Northern.tech/Mender/useradm-enterprise/-/pipelines/1287885415

PR #279, opened 2024-05-17
allow overriding name and fullname

ignatiusreza:
we're packaging mender as a sub-chart. which resulted in a relatively long chart names combination:

```
Invalid value: `our-custom-helm-chart-mender-deployments-storage-daemon` must be no more than 52 characters
```

### DIFF
--- a/mender/values.yaml
+++ b/mender/values.yaml
@@ -1,3 +1,5 @@
+fullnameOverride: ""
+
 global:
   enterprise: true
   hosted: false
--- a/README.md
+++ b/README.md
@@ -252,6 +252,7 @@ The following table lists the global, default, and other parameters supported by
 | `serviceAccount.name` | Custom ServiceAccount name | `nil` |
 | `serviceAccount.labels` | ServiceAccount labels | `nil` |
 | `serviceAccount.annotations` | ServiceAccount annotations | `nil` |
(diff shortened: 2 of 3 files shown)


mender/values.yaml
```diff
@@ -1,3 +1,6 @@
+nameOverride: ""
```
oldgiova, line 1: Can't understand this one, it seems not used to me

oldgiova (commented):
Thanks for your contribution! It looks like to me, only a few nitpicks: 

Can you please amend your commit and add a signoff (`-s` parameter)? See our [contributing doc](https://github.com/mendersoftware/mender/blob/master/CONTRIBUTING.md#providing-pull-requests). Sorry for being pedantic, but this is required...

Can you also add this change to the README? Something like this at line 255
```
| `fullnameOverride` | Use a custom prefix for the resources, instead of using the helm chart name (default) | `nil` |
```

mender/values.yaml
ignatiusreza: > https://github.com/mendersoftware/mender-helm/blob/master/mender/templates/_helpers.tpl#L6

It's actually already being used, but it's undocumented.. I thought, it makes sense to add it to make it more visible..

I removed it from the commit for now, and maybe it can be addressed separately.

oldgiova: You're right! Thanks for bringing this up, I'll add this in a later PR

oldgiova (approved):

PR #285, opened 2024-05-22
Adding dependabot

oldgiova:
Ticket: SEC-1174


alfrunes (commented):
Maybe a silly question, but what exactly is this doing? There are no `Dockerfile` in the `/mender` directory.

alfrunes (approved):
Let's try and see what happens

PR #288, opened 2024-05-24
Added tenantadm and iotmanager CronJobs

oldgiova:
This is an Hosted only feature: cronjobs to maintain tenants

Ticket: MC-7433

Hosted Mender alignements:
* `cronjob-mender-tenantadm-suspend-expired.yaml` originally contains two containers: this is splitted into two different CronJobs with the same schedules and the two different containers: this for reusing the _podtemplate
* `cronjob-mender-tenantadm-sync-tenants.yaml`
* `cronjob-iot-manager-sync-devices.yaml`

### DIFF
--- a/README.md
+++ b/README.md
@@ -649,6 +649,13 @@ The following table lists the parameters for the `tenantadm` component and their
 | `tenantadm.probesOverrides.successThreshold` | Override the `successThreshold` for every Readiness and Liveness probes. | `nil` |
 | `tenantadm.probesOverrides.timeoutSeconds` | Override the `timeoutSeconds` for every Readiness and Liveness probes. | `nil` |
 | `tenantadm.probesOverrides.failureThreshold` | Override the `failureThreshold` for every Readiness and Liveness probes. | `nil` |
+| `tenantadm.cronjobs.suspendTenants.enabled` | Creates the Suspend Tenants cronjob (Hosted only) | `false` |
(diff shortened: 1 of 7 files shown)


oldgiova:
> I'm not sure about exposing internal tools to the public Helm chart. I believe the same also applies for other tools not present in this change. Could we maybe instead create a separate helm chart or kustomize environment for managing these resources?

I wasn't sure either, but then I found this one: https://github.com/mendersoftware/mender-helm/blob/master/mender/templates/device-auth/cronjob.yaml  :) 

The issue is that the source of truth of the `tenantadm` image is now in the mender-helm repo... But let's talk about this:
1. Separate helm chart: to inherit the same `tenantadm` version, that would be a subchart... But this subchart could not be private, so that would be a new public helm chart, so the principle it's the same of this PR, but it's an overcomplication
2. kustomize: then we have to manually patch the `tenantadm` image, and I'm pretty sure we'll miss some update.. 
3. New terraform modules: same as 2.

Maybe, we could use either kustomize or a new terraform module, and using the `tenantadm:master` image? If we really don't want to make these cronjobs public. But imho, since there's the device-auth cronjob already public, I see no value in moving these away from the source of truth (not a strong opinion btw)

oldgiova:
Superseded by https://github.com/mendersoftware/mender-helm/pull/423

PR #298, opened 2024-05-30
v5.8.0

oldgiova:
* Removing helm chart label
* Added Helm chart tests
* Added deployments service' redis connection

### DIFF
--- a/mender/templates/deployments/_podtemplate.yaml
+++ b/mender/templates/deployments/_podtemplate.yaml
@@ -81,6 +81,7 @@ spec:
     env:
     - name: DEPLOYMENTS_STORAGE_DEFAULT
       value: {{ .dot.Values.global.storage | quote }}
+
     - name: DEPLOYMENTS_MIDDLEWARE
       value: {{ .dot.Values.deployments.env.DEPLOYMENTS_MIDDLEWARE | quote }}
     - name: DEPLOYMENTS_AWS_TAG_ARTIFACT
@@ -101,6 +102,14 @@ spec:
       value: "true"
     {{- end }}
     {{- end }}
+
+    {{- if .dot.Values.global.enterprise }}
+    - name: DEPLOYMENTS_REDIS_CONNECTION_STRING
+      value: {{ include "redis_connection_string" . }}
+    - name: DEPLOYMENTS_REDIS_KEY_PREFIX
+      value: {{ .dot.Values.device_auth.env.DEPLOYMENTS_REDIS_KEY_PREFIX | default "deployments:v1" | quote }}
+    {{- end }}
+
     {{- include "mender.customEnvs" (merge (deepCopy .dot.Values.deployments) (deepCopy (default (dict) .dot.Values.default))) | nindent 4 }}
 
     # Supported configuration settings: https://github.com/mendersoftware/deployments/blob/master/config.yaml
--- a/mender/templates/tests/test-mender.yaml
+++ b/mender/templates/tests/test-mender.yaml
@@ -30,11 +30,11 @@ spec:
           echo "Testing deployment service";
           wget -qO- '{{ include "mender.fullname" . }}-deployments:{{ .Values.deployments.service.port }}/api/internal/v1/deployments/health';
           {{- end }}
-          {{- if .Values.deviceauth.enabled }}
+          {{- if .Values.device_auth.enabled }}
           echo "Testing deviceauth service";
           wget -qO- '{{ include "mender.fullname" . }}-device-auth:{{ .Values.device_auth.service.port }}/api/internal/v1/devauth/health';
           {{- end }}
-          {{- if .Valued.deviceconfig.enabled }}
+          {{- if .Values.deviceconfig.enabled }}
           echo "Testing deviceconfig service";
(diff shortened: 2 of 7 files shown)


mender/templates/tests/test-mender.yaml
```diff
@@ -0,0 +1,46 @@
...
+          echo "Testing deviceconfig service";
+          wget -qO- '{{ include "mender.fullname" . }}-deviceconfig:{{ .Values.deviceconfig.service.port }}/api/internal/v1/deviceconfig/health';
+          echo "Testing deviceconnect service";
+          wget -qO- '{{ include "mender.fullname" . }}-deviceconnect:{{ .Values.deviceconnect.service.port }}/api/internal/v1/deviceconnect/health';
+          echo "Testing inventory service";
+          wget -qO- '{{ include "mender.fullname" . }}-inventory:{{ .Values.inventory.service.port }}/api/internal/v1/inventory/health';
+          echo "Testing iot_manager service";
+          wget -qO- '{{ include "mender.fullname" . }}-iot-manager:{{ .Values.iot_manager.service.port }}/api/internal/v1/iot-manager/health';
+          echo "Testing useradm service";
+          wget -qO- '{{ include "mender.fullname" . }}-useradm:{{ .Values.useradm.service.port }}/api/internal/v1/useradm/users';
```
alfrunes, line 35: Why not use the `/health` endpoint?
```suggestion
          wget -qO- '{{ include "mender.fullname" . }}-useradm:{{ .Values.useradm.service.port }}/api/internal/v1/health';
```

mender/templates/tests/test-mender.yaml
```diff
@@ -0,0 +1,46 @@
...
+spec:
+  containers:
+    - name: wget
+      image: busybox
+      command: ['/bin/sh', '-c']
+      args:
+        {{- if and (.Values.global.enterprise) (.Values.tenantadm.enabled) (.Values.auditlogs.enabled) }}
+        - echo "Starting smoketests";
+          echo "Testing deployment service";
+          wget -qO- '{{ include "mender.fullname" . }}-deployments:{{ .Values.deployments.service.port }}/api/internal/v1/deployments/health';
```
alfrunes, line 19: Strictly speaking, shouldn't there be a check
```
{{ if .Values.deployments.enabled }}
...
```
Although I'm not sure if it makes much sense running without the services.

mender/templates/tests/test-mender.yaml
```diff
@@ -0,0 +1,46 @@
...
+    {{- include "mender.labels" . | nindent 4 }}
+  annotations:
+    "helm.sh/hook": test
+spec:
+  containers:
+    - name: wget
+      image: busybox
+      command: ['/bin/sh', '-c']
+      args:
+        {{- if and (.Values.global.enterprise) (.Values.tenantadm.enabled) (.Values.auditlogs.enabled) }}
```
alfrunes, line 16: This condition is a bit weird, since `.Values.global.enterprise` only adds `auditlogs`, `devicemonitor` and `tenantadm`. The rest of the stack should be the same.

mender/templates/tests/test-mender.yaml
oldgiova: Because the health endpoint is actually already checked by the health probe. The `/useradm/users` is an "application check", so it's even better than the health, in the smoke test scope

mender/templates/tests/test-mender.yaml
oldgiova: I think we have to restructure the `enabled` services...

mender/templates/tests/test-mender.yaml
alfrunes: The same applies to tests for the other services. OTOH, there is some logic behind the `/health` endpoints (it pings the databases and other depending services).

mender/templates/tests/test-mender.yaml
alfrunes: Figured it was easier to submit a PR than explain it. Here you go: https://github.com/oldgiova/mender-helm/pull/1 :)

mender/templates/deployments/_podtemplate.yaml
```diff
@@ -101,6 +102,33 @@ spec:
...
+    # redis_address deprecated from 3.7
+    - name: DEPLOYMENTS_REDIS_CONNECTION_STRING
+      value: {{ include "redis_connection_string" . }}
+    {{- end }}
+    - name: DEPLOYMENTS_REDIS_KEY_PREFIX
+      value: {{ .dot.Values.device_auth.env.DEPLOYMENTS_REDIS_KEY_PREFIX | default "deployments:v1" | quote }}
+    {{- end }}
+    - name: DEPLOYMENTS_REDIS_LIMITS_EXPIRE_SEC
+      value: {{ .dot.Values.device_auth.env.DEPLOYMENTS_REDIS_LIMITS_EXPIRE_SEC | default "3600" | quote }}
+    {{- end }}
```
alfrunes, line 130: There are only two `REDIS_` configuration variables for deployments
```suggestion
    - name: DEPLOYMENTS_REDIS_CONNECTION_STRING
      value: {{ include "redis_connection_string" . }}
    - name: DEPLOYMENTS_REDIS_KEY_PREFIX
      value: {{ .dot.Values.device_auth.env.DEPLOYMENTS_REDIS_KEY_PREFIX | default "deployments:v1" | quote }}
```

alfrunes (approved):

PR #310, opened 2024-06-14
Updgrade to 3.7.5.

kacf:


kacf:
@oldgiova: The upgrade tests in the gitlab-ci file might be a bit redundant now. They are:

* `helm_chart_upgrade_from_previous_lts_to_current_lts`
* `helm_chart_upgrade_from_previous_release_to_current_lts`
* `helm_chart_upgrade_from_current_lts_to_current_release`

However, the previous release, 3.4, was EOL already a long time ago, so we just have the two LTSes now. Maybe we can remove at least one of those tests.

Also, there is a changelog which I wasn't quite sure how to update.

oldgiova:
> @oldgiova: The upgrade tests in the gitlab-ci file might be a bit redundant now. They are:
> 
> * `helm_chart_upgrade_from_previous_lts_to_current_lts`
> * `helm_chart_upgrade_from_previous_release_to_current_lts`
> * `helm_chart_upgrade_from_current_lts_to_current_release`
> 
> However, the previous release, 3.4, was EOL already a long time ago, so we just have the two LTSes now. Maybe we can remove at least one of those tests.
> 
> Also, there is a changelog which I wasn't quite sure how to update.

The changelog is more for the customers, you're change is internal only so don't worry about it.

As for the tests, now the latest release is LTS, but the tests should work also when the latest release is not LTS, so they still make sense to me.

.gitlab-ci.yml
```diff
@@ -326,15 +326,15 @@ test:helm_chart_upgrade_from_previous_lts_to_current_lts:
   script:
     - tests/ci-make-deps.sh
     - |
-      echo "INFO - installing mender tag ${MENDER_PREVIOUS_LTS_RELEASE_TAG:-mender-3.3.2} with helm chart version ${MENDER_LTS_HELM_VERSION:-5.0.1}"
+      echo "INFO - installing mender tag ${MENDER_PREVIOUS_LTS_RELEASE_TAG:-mender-3.6.5} with helm chart version ${MENDER_LTS_HELM_VERSION:-5.0.1}"
```
oldgiova, line 329: Could you also switch to the latest helm chart? Currently it's `5.9.0`. This because the previous LTS was compatible up to the 5.0 only, but now the previous LTS (`3.6`) is compatible with the latest helm chart.
```suggestion
      echo "INFO - installing mender tag ${MENDER_PREVIOUS_LTS_RELEASE_TAG:-mender-3.6.5} with helm chart version ${MENDER_LTS_HELM_VERSION:-5.9.0}"
```

.gitlab-ci.yml
```diff
@@ -343,14 +343,14 @@ test:helm_chart_upgrade_from_previous_lts_to_current_lts:
         mender/mender --version ${MENDER_LTS_HELM_VERSION:-5.0.1} || exit 3;
```
oldgiova, line 343: Here as well
```suggestion
        mender/mender --version ${MENDER_LTS_HELM_VERSION:-5.9.0} || exit 3;
```

.gitlab-ci.yml
```diff
@@ -471,15 +471,15 @@ test:helm_chart_upgrade_from_current_lts_to_current_release:
   script:
     - tests/ci-make-deps.sh
     - |
-      echo "INFO - installing mender tag ${MENDER_CURRENT_LTS_RELEASE_TAG:-mender-3.6.3}"
+      echo "INFO - installing mender tag ${MENDER_CURRENT_LTS_RELEASE_TAG:-mender-3.6.5}"
```
oldgiova, line 474: This is supposed to be the current LTS:
```suggestion
      echo "INFO - installing mender tag ${MENDER_CURRENT_LTS_RELEASE_TAG:-mender-3.7.5}"
```
This is for testing the upgrade from an already release helm chart (chart.mender.io) to the local helm chart.

.gitlab-ci.yml
```diff
@@ -471,15 +471,15 @@ test:helm_chart_upgrade_from_current_lts_to_current_release:
...
       source ./tests/variables.sh
       helm upgrade -i mender \
         -f mender/values.yaml \
         -f tests/keys.yaml \
         -f tests/values-helmci.yaml \
         --wait \
         --timeout=${HELM_UPGRADE_TIMEOUT:-15m} \
-        --set global.image.tag="${MENDER_CURRENT_LTS_RELEASE_TAG:-mender-3.6.3}" \
+        --set global.image.tag="${MENDER_CURRENT_LTS_RELEASE_TAG:-mender-3.6.5}" \
```
oldgiova, line 482: Same as above
```suggestion
        --set global.image.tag="${MENDER_CURRENT_LTS_RELEASE_TAG:-mender-3.7.5}" \
```

mender/Chart.yaml
```diff
@@ -1,5 +1,5 @@
 apiVersion: v2
-appVersion: "3.7.4"
+appVersion: "3.7.5"
 description: Mender is a robust and secure way to update all your software and deploy your IoT devices at scale with support for customization
 name: mender
 version: 5.9.0
```
oldgiova, line 5: When we create a new appVersion release, we also bump the helm chart version:
```suggestion
version: 5.9.1
```

Actually, it deserves a mention in the CHANGELOG: something like this:

```
## Version 5.9.1
* Upgrade to Mender version `3.7.5`
```

oldgiova (commented):
Also:

## test:helm_chart_upgrade_from_previous_release_to_current_lts:
line 401:
```
# fom:
echo "INFO - installing mender tag ${MENDER_PREVIOUS_RELEASE_TAG:-mender-3.4.0} with helm chart version ${MENDER_LTS_HELM_VERSION:-5.0.1}"
# to:
echo "INFO - installing mender tag ${MENDER_PREVIOUS_RELEASE_TAG:-mender-3.6.5} with helm chart version ${MENDER_LTS_HELM_VERSION:-5.9.0}"
```
line 409:
```
# fom:
--set global.image.tag="${MENDER_PREVIOUS_RELEASE_TAG:-mender-3.4.0}" \
# to:
--set global.image.tag="${MENDER_PREVIOUS_RELEASE_TAG:-mender-3.6.5}" \
```

Line 415:
```
# from:
mender/mender --version ${MENDER_LTS_HELM_VERSION:-5.0.1} || exit 3;
# to:
mender/mender --version ${MENDER_LTS_HELM_VERSION:-5.9.0} || exit 3;
```

oldgiova:
> The changelog is more for the customers, you're change is internal only so don't worry about it.

I was wrong: could you please add this in the CHANGELOG.md?
```
## Version 5.9.1
* Upgrade to Mender version `3.7.5`
```
Thanks!

.gitlab-ci.yml
kacf: But then both the old and the new will be 3.7.5, so it will be upgrading to the same version it came from. Is that correct?

kacf:
> As for the tests, now the latest release is LTS, but the tests should work also when the latest release is not LTS, so they still make sense to me.

Sure. But changing away from 3.4.0 makes the two jobs `helm_chart_upgrade_from_previous_lts_to_current_lts` and `helm_chart_upgrade_from_previous_release_to_current_lts` identical, AFAICS.

.gitlab-ci.yml
oldgiova: They are the same Mender version (this time, not the next non-LTS release), but most importantly, they are two different Helm Chart versions: here the test is focusing on the Helm code, not on Mender versions:

Line 488: the first installation installs the current LTS with the latest charts.mender.io chart (`--version` is not specified, so it get the latest`): 
```
        mender/mender || exit 3;
```
Line 504: the test upgrades to the latest Helm code, present in `master`, which is still not published in the charts.mender.io; mender will be the latest release (LTS or not)
```
        ./mender || exit 3;
```

oldgiova:
> > > As for the tests, now the latest release is LTS, but the tests should work also when the latest release is not LTS, so they still make sense to me.
> > 
> > 
> > Sure. But changing away from 3.4.0 makes the two jobs `helm_chart_upgrade_from_previous_lts_to_current_lts` and `helm_chart_upgrade_from_previous_release_to_current_lts` identical, AFAICS.
> 
> That's true for this release, but not when the latest release will be `mender-3.8.0`, non LTS.

Actually, double thinking, I think it could still have sense to test the upgrade from 3.4.0 (and helm 5.0) to the 3.7.5, just in case...

.gitlab-ci.yml
kacf: Aha, that makes sense!

kacf:
Sure. So just to recap and make sure I've done it correctly:

* Revert the `helm_chart_upgrade_from_previous_release_to_current_lts` job back to upgrading from 3.4.0.
* In the `helm_chart_upgrade_from_current_lts_to_current_release` job, switch to upgrade from Mender 3.7.5 to 3.7.5. It is really the Helm chart version which is being upgrade-tested, not Mender.

Sounds good?

oldgiova:
> Sure. So just to recap and make sure I've done it correctly:
> 
> * Revert the `helm_chart_upgrade_from_previous_release_to_current_lts` job back to upgrading from 3.4.0.
> * In the `helm_chart_upgrade_from_current_lts_to_current_release` job, switch to upgrade from Mender 3.7.5 to 3.7.5. It is really the Helm chart version which is being upgrade-tested, not Mender.
> 
> Sounds good?

Yup!

oldgiova:
Sorry, last thing: also the README.md, line 203, need to be bumped to version 3.7.5

kacf:
Thanks, I missed that. Fixed!

oldgiova (approved):

PR #338, opened 2024-08-21
fix(traefik v3): Invalid host regex for default storage proxy URL

alfrunes:
The regular expression in the hostname rule was using incompatible Traefik v2 regex.


oldgiova (approved):

oldgiova:
I think this requires also a new version bump, for the on-prem releases mostly.

alfrunes:
Updated CHANGELOG.md and bumped the Chart version.

PR #341, opened 2024-09-11
Release please CHANGELOG automation

oldgiova:
Added release-please and git cliff. After including a new `fix|feat|feat!` to `master`, the tools are creating a new Changelog PR with the new version bumped accordingly to the commit messages.

**Notes**
1. the old changelog will be overridden 
2. the new tag should include the component: `mender-5.10.2` instead of `5.10.2`. This is configurable, but I'm doing this in preparation to the possible `mender-server` component


.gitlab-ci.yml
```diff
@@ -790,3 +817,77 @@ cleanup_eks_cluster:success:
...
+        && git cliff --output mender/CHANGELOG.md --bump --github-repo ${GITHUB_REPO_URL}
+        && git add mender/CHANGELOG.md
+        && git commit --amend -s --no-edit
+        && git push origin --force
+    # note: using sed to prevent release-please to mess up with the Github Release body
+    - echo "INFO - updating PR body"
+        && git cliff --unreleased --bump -o tmp_pr_body.md --github-repo ${GITHUB_REPO_URL}
+        && sed -i 's/## mender-\([0-9.]*\)/## \1/' tmp_pr_body.md
+        && gh pr edit $RELEASE_PLEASE_PR --body-file tmp_pr_body.md
+        && rm tmp_pr_body.md
```
alfrunes, line 876: Every time I read this script I keep wondering what's the purpose of release-please. Isn't it simply a glorified `gh pr create`/`gh release create`?
If I'm not mistaken, this snippet is forcefully overwriting the entire content of the PR.

alfrunes (approved):

.gitlab-ci.yml
oldgiova: Added the `--unreleased` flag to just add the latest release, without overriding the previous ones.

> Every time I read this script I keep wondering what's the purpose of release-please. Isn't it simply a glorified gh pr create/gh release create?

That, but most importantly it has the logic to check every commit and understand which will be the next version; the logic also applies to maintenance branches. 
We could just rewrite this logic implementing a git cliff wrapper in bash/python, but this will end up in a custom tool, something that we'd like to avoid. The tool could be pretty simple, but would still require a fair amount of code and cases.

OTOH, we're still considering release-please experimental, so if the experience is these "testing" repositories is bad, we could simply try with another (internal?) tool.

PR #347, opened 2024-10-07
ci: fix changelog append on top

oldgiova:
With the --unreleased flag, only the last release were added to the CHANGELOG.md, so the old history would have been lost. With this change, we'll just prepend the new git cliff release.

[Wrong CHANGELOG patch](https://github.com/mendersoftware/mender-helm/pull/346/commits/da9a4ef2e668fa0fa057321e0aac287b01f38997#diff-d2cbe88096a4c24cd4a33d862fe49df122c543515c8b811b79f67818f655982e)

Likely good CHANGELOG patch:
```
# Mender Helm Chart


## 5.11.0 - 2024-10-07


### Features


- Add gui hpa ([8f6d9f4](https://github.com/mendersoftware/mender-helm.git/commit/8f6d9f46c0c9db16939a8c851ed4bda21e7ca5f3)) 
  Added Horizontal Pod Autoscaler resource to the gui container, to scale
  it automatically when the service experiences more load.

## Version 5.10.1
* Fix invalid regexp in default storage proxy rule.

## Version 5.10.0
* Change `generate_delta_worker` to StatefulSet and add `persistence` values
  * The new `persistence` values specifies the parameters of the PVC template of
    the statefulset.
* Upgrade Traefik to v3.1.2
* Upgrade to Mender version `3.7.7`
```

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -858,13 +858,15 @@ changelog:
         --target-branch=${CI_COMMIT_REF_NAME}
     # git cliff: override the changelog
     - test $GIT_CLIFF == "false" && echo "INFO - Skipping git-cliff" && exit 0
-    - git clone https://${GITHUB_USER}:${GITHUB_BOT_TOKEN_REPO_FULL}@github.com/${GITHUB_REPO_URL}
-    - cd ${GITHUB_REPO_URL#*/}
+    - git remote add github https://${GITHUB_USER}:${GITHUB_BOT_TOKEN_REPO_FULL}@github.com/${GITHUB_REPO_URL}
+    - gh repo set-default https://${GITHUB_USER}:${GITHUB_BOT_TOKEN_REPO_FULL}@github.com/${GITHUB_REPO_URL}
     - RELEASE_PLEASE_PR=$(gh pr list --author "${GITHUB_USER_NAME}" --head "release-please--branches--${CI_COMMIT_REF_NAME}" --json number | jq -r '.[0].number // empty')
     - test -z "$RELEASE_PLEASE_PR" && echo "No release-please PR found" && exit 0
     - echo "INFO - generating changelog notes from git cliff"
         && gh pr checkout $RELEASE_PLEASE_PR
-        && git cliff --unreleased --output mender/CHANGELOG.md --bump --github-repo ${GITHUB_REPO_URL}
+        && git cliff --unreleased --bump --prepend mender/CHANGELOG.md --github-repo ${GITHUB_REPO_URL}
+        && sed -i 's/# Mender Helm chart//' mender/CHANGELOG.md
+        && sed -i 's/---/# Mender Helm chart\n/' mender/CHANGELOG.md
         && git add mender/CHANGELOG.md
         && git commit --amend -s --no-edit
         && git push origin --force


oldgiova:
Of course, an alternative is just to rewrite the full changelog every time, without the `--unreleased` flag. Theoretically, the old releases should not change

.gitlab-ci.yml
```diff
@@ -858,13 +861,16 @@ changelog:
...
     - RELEASE_PLEASE_PR=$(gh pr list --author "${GITHUB_USER_NAME}" --head "release-please--branches--${CI_COMMIT_REF_NAME}" --json number | jq -r '.[0].number // empty')
+    - gh repo set-default https://${GITHUB_USER}:${GITHUB_BOT_TOKEN_REPO_FULL}@github.com/${GITHUB_REPO_URL}
     - test -z "$RELEASE_PLEASE_PR" && echo "No release-please PR found" && exit 0
     - echo "INFO - generating changelog notes from git cliff"
         && gh pr checkout $RELEASE_PLEASE_PR
-        && git cliff --unreleased --output mender/CHANGELOG.md --bump --github-repo ${GITHUB_REPO_URL}
+        && new_changelog=$(git cliff --unreleased --bump --github-repo ${GITHUB_REPO_URL} | sed 's/---//')
+        && echo $new_changelog | cat - mender/CHANGELOG.md > tmp_changelog.md
+        && echo "${CHANGELOG_HEADER}\n" > mender/CHANGELOG.md
+        && cat tmp_changelog.md | grep -v ${CHANGELOG_HEADER} >> mender/CHANGELOG.md
```
alfrunes, line 873: What about using `--prepend` instead of `--output`?
```suggestion
        && git cliff --unreleased --prepend mender/CHANGELOG.md --bump --github-repo ${GITHUB_REPO_URL}
```

oldgiova: Yeah! Much nicer!

oldgiova: I still have to update the header `# Mender Helm chart` and remove the `---`, but it's better than before.

.gitlab-ci.yml
```diff
@@ -858,13 +858,15 @@ changelog:
...
+    - git remote add github https://${GITHUB_USER}:${GITHUB_BOT_TOKEN_REPO_FULL}@github.com/${GITHUB_REPO_URL}
     - RELEASE_PLEASE_PR=$(gh pr list --author "${GITHUB_USER_NAME}" --head "release-please--branches--${CI_COMMIT_REF_NAME}" --json number | jq -r '.[0].number // empty')
+    - gh repo set-default https://${GITHUB_USER}:${GITHUB_BOT_TOKEN_REPO_FULL}@github.com/${GITHUB_REPO_URL}
     - test -z "$RELEASE_PLEASE_PR" && echo "No release-please PR found" && exit 0
     - echo "INFO - generating changelog notes from git cliff"
         && gh pr checkout $RELEASE_PLEASE_PR
-        && git cliff --unreleased --output mender/CHANGELOG.md --bump --github-repo ${GITHUB_REPO_URL}
+        && git cliff --unreleased --bump --prepend mender/CHANGELOG.md --github-repo ${GITHUB_REPO_URL}
+        && sed -i 's/# Mender Helm chart//' mender/CHANGELOG.md
+        && sed -i 's/---/# Mender Helm chart\n/' mender/CHANGELOG.md
```
alfrunes, line 869: This seems to be a side effect of poor choice of changelog header.

alfrunes (approved):
I do not understand why we insist on keeping this template shared in a remote repository. The changelog format for a helm chart is going to be different from a changelogs generated for terraform modules which in turn is different from a changelog generated for a standalone go-component which again is going to be different from a multi-project repository. To me this seems to just add more complexity than what we gain from it (moreover, all those changelogs are suddenly going to change once we change the shared template).

Anyway, if we decide to keep working around it I won't block it.

mzedel (approved):
🕺 Why do we even keep the helm chart and the rest in different places? The changelog for the frontend is going to be different from the useradm service - why would we keep these in the same repo, but not the helm chart? To me keeping these in different places seems to just add more complexity...

Frontend and helm chart both consume artifacts (images, API responses) still we somehow make the distinction (something similar could surely be found for other services)... we distribute both alongside each other and they are not interchangeable (say run 3.7 in the next helm version, or even 2.7).

oldgiova:
> I do not understand why we insist on keeping this template shared in a remote repository.
There's no clear agreement inside the team; I personally prefer using the template from the internal repository, but for the app-update-module project, it was chosen a centralized one because we wanted to always reuse a single standard (or two: standard and scoped).
I think a decision on how the changelog will be is far for being consolidated, so I'm just considering this an evolving process, starting from this point.

> Why do we even keep the helm chart and the rest in different places?
It has been a team decision at the very beginning of the monorepo project; I asked to merge the helm-chart in the monorepo, the team choice was to let it standalone. Thus, we could talk about it again, imho

oldgiova:
Force-pushed a small fix: moved `gh repo set-default https://${GITHUB_USER}:${GITHUB_BOT_TOKEN_REPO_FULL}@github.com/${GITHUB_REPO_URL}` before the `gh pr list` command.

PR #350, opened 2024-10-10
ci: fix diverging PRs

oldgiova:
In some conditions the upstream pr branch could diverge from the local branch. With this fix it should rebase the upstream changes locally.
For example: https://gitlab.com/Northern.tech/Mender/mender-helm/-/jobs/8049202233#L251

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -863,6 +863,8 @@ changelog:
     - test -z "$RELEASE_PLEASE_PR" && echo "No release-please PR found" && exit 0
     - echo "INFO - generating changelog notes from git cliff"
         && gh pr checkout $RELEASE_PLEASE_PR
+        && git fetch github-${TS_SUFFIX} pull/${RELEASE_PLEASE_PR}/head
+        && git reset --hard FETCH_HEAD
         && wget --output-document cliff.toml https://raw.githubusercontent.com/mendersoftware/mendertesting/master/utils/cliff.toml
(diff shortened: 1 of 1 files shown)


oldgiova:
A possible alternative is the original behavior: just clone from scratch the github repo into a temp folder every time, and add modification into this, instead using the preexisting local `.git`.

.gitlab-ci.yml
```diff
@@ -863,6 +863,7 @@ changelog:
     - test -z "$RELEASE_PLEASE_PR" && echo "No release-please PR found" && exit 0
     - echo "INFO - generating changelog notes from git cliff"
         && gh pr checkout $RELEASE_PLEASE_PR
+        && git pull
```
alfrunes, line 866: Couldn't this potentially cause merge conflicts?
What about a simple `git fetch github-${TS_SUFFIX} pull/$RELEASE_PLEASE_PR/head && git reset --hard FETCH_HEAD`?

oldgiova: yeah, thanks for the trick!

alfrunes (approved):
Just a side comment to #348; maybe it would be cleaner to simply ignore the error when adding the remote instead of carrying around the `TS_SUFFIX` env variable.
```bash
git remote add github https://${GITHUB_USER}:${GITHUB_BOT_TOKEN_REPO_FULL}@github.com/${GITHUB_REPO_URL} || true
```

PR #351, opened 2024-10-10
ci: unique remote github repo

oldgiova:
@alfrunes I saw that in the same runner, sometime old github remote could "survive", and could be from past try and tests, so there is no guarantees the `github` remote is the good one.
This PR address this behavior by always setting a new unique github remote `github-${CI_JOB_ID}`, and removes it in the `after_script`. Successfully tested in a [sample pipeline](https://gitlab.com/Northern.tech/northerntechhq/nt-boilerplate-pipeline/-/jobs/8070161234)

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -834,18 +834,15 @@ changelog:
     # install release-please
     - npm install -g release-please
     # install github-cli
-    - |
-      mkdir -p -m 755 /etc/apt/keyrings \
-        && wget -qO- https://cli.github.com/packages/githubcli-archive-keyring.gpg | tee /etc/apt/keyrings/githubcli-archive-keyring.gpg > /dev/null \
-        && chmod go+r /etc/apt/keyrings/githubcli-archive-keyring.gpg \
-        && echo "deb [arch=$(dpkg --print-architecture) signed-by=/etc/apt/keyrings/githubcli-archive-keyring.gpg] https://cli.github.com/packages stable main" | tee /etc/apt/sources.list.d/github-cli.list > /dev/null \
-        && apt update \
-        && apt install gh jq -y
-    # Setting up git
-    - |
-      echo "INFO - setting up git"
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -856,25 +856,27 @@ changelog:
...
-        && git reset --hard FETCH_HEAD
-        && wget --output-document cliff.toml https://raw.githubusercontent.com/mendersoftware/mendertesting/master/utils/cliff.toml
+    - git fetch github-${CI_JOB_ID} pull/${RELEASE_PLEASE_PR}/head
+    - git reset --hard FETCH_HEAD
+    - wget --output-document cliff.toml https://raw.githubusercontent.com/mendersoftware/mendertesting/master/utils/cliff.toml
         && git cliff --bump --output mender/CHANGELOG.md --github-repo ${GITHUB_REPO_URL}
         && git add mender/CHANGELOG.md
         && git commit --amend -s --no-edit
-        && git push github --force
+        && git push github-${CI_JOB_ID} --force
```
alfrunes, line 871: Could we replace all `&&` with `-`?
Logically it's the same, only the latter is much easier to debug. Instead of seeing the first command and "collapsed multiline command" on failure, you will see the actual command that failed.

alfrunes (commented):
I still don't quite get it; it's the same remote. Even if the local git repository is cached, `git fetch` still fetches the upstream PR.

.gitlab-ci.yml
oldgiova: The reason for all this `&&` is because you have a size limit of the array script; if I good remember is 10, but let me try again.

oldgiova:
> I still don't quite get it; it's the same remote. Even if the local git repository is cached, `git fetch` still fetches the upstream PR.

From my tests, I found an old remote from a former test with the same name, but incorrectly formatted. So we have no 100% guarantees that the remote is actually the same/

PR #360, opened 2024-10-31
ci: fix seaweedfs filer data storage

oldgiova:
For some reason, the data storage is not working with the default PVC and StorageClass. Falling back to emptyDir, is enough for the tests

Ticket: None
Changelog: None

### DIFF
--- a/tests/seaweedfs.yaml
+++ b/tests/seaweedfs.yaml
@@ -7,7 +7,8 @@ filer:
     enableAuth: true
     createBuckets:
       - name: "mender-artifacts-seaweedfs"
-  storageClass: "gp2"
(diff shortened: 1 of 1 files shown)


alfrunes (approved):
I suppose emptyDir is good enough for tests, but it might be worthwhile to investigate this later since it's used in the docs.

oldgiova:
> I suppose emptyDir is good enough for tests, but it might be worthwhile to investigate this later since it's used in the docs.

In the docs I used k3s, and persistence was working. I guess (hope) that no ones is using Seaweedfs while using EKS... But yeah, I'm curious about it is failing

PR #386, opened 2024-11-21
Align prod with master :warning: wait from the release day

oldgiova:


alfrunes:
I can see #398 and #394 is missing. Should we also include them here?

oldgiova:
> I can see #398 and #394 is missing. Should we also include them here?

Yes, we should, but #398 requires https://github.com/mendersoftware/mender-helm/pull/401

alfrunes (approved):

PR #398, opened 2024-12-03
refac: consistency between nats_uri declaration

oldgiova:
Aligned the nats_uri of the deviceconnect and the workflows definitions to the other services

Ticket: None


alfrunes (approved):
Thanks for cleaning this up :+1:

PR #399, opened 2024-12-04
feat: Upgrade to Mender version `3.7.8`

lluiscampos:


oldgiova (changes requested):
We could do this, but with the `5.x` branch as a target

oldgiova (approved):

PR #407, opened 2025-01-07
fix!: remove errorPort from gui

oldgiova:
Follow up removing the error reponder middleware from Traefik

Ticket: None


alfrunes:
You also need to remove the json-error-responder middlewares (including the attachment to routes).

oldgiova:
> You also need to remove the json-error-responder middlewares (including the attachment to routes).

@alfrunes done, thanks for the suggestion

PR #417, opened 2025-01-13
Add missing tenantadm env

alfrunes:


oldgiova (approved):
Does it has effects on hM?

alfrunes:
> Does it has effects on hM?

No, it is not. The default values are correct. It's only if you modify the `*.service.name` or `*.service.port` values that you may get into trouble. The missing environment variables are only relevant for the upcoming 4.0 release.

oldgiova (approved):

PR #421, opened 2025-01-17
chore: Workaround for SeaweedFS through storage proxy

alfrunes:
Drops X-Forwarded-Host and X-Forwarded-Port when
`storage_proxy.enabled`.


oldgiova (approved):
This solves the issue for the deployments service, which is an improvement. The create-artifacts-worker is still failing, but this is another story.
Approving for now!

PR #439, opened 2025-02-18
Optionals deviceauth cronjobs

oldgiova:
You can create custom cronjobs based on the deviceauth service. This feature is going to deprecate the device_license_count feature, that could be installed as a custom cronjob, with the provided example in the Values file.
The device_license_count key will be removed in the future.

Ticket: MC-7789


merlin-northern:
it was created to pull https://github.com/mendersoftware/saas/pull/1417 off thanks Roberto!

merlin-northern (approved):
thank you!

alfrunes (approved):

PR #443, opened 2025-02-25
Switch api-gateway tls certificate to k8s tls type

chriswiggins:
We currently have nginx sitting in front of the api-gateway, doing SSL termination. It is also doing storage proxying, but we can move that to Traefik now with the latest release 😄 

We use cert-manager for our certificates, and currently we can't pass a secret cert-manager creates into the api-gateway, as it uses the wrong keys.

We prefer not to use k8's ingress, for a couple of reasons, so want to have Traefik serve the traffic directly

### DIFF
--- a/mender/templates/api-gateway/configmap.yaml
+++ b/mender/templates/api-gateway/configmap.yaml
@@ -547,8 +547,8 @@ data:
 {{- if $isTls }}
     tls:
       certificates:
-        - certFile: /etc/traefik/certs/cert.crt
-          keyFile: /etc/traefik/certs/private.key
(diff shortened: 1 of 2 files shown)


oldgiova:
Hi @chriswiggins thank you so much for your contribution! I'm reviewing and testing it.

oldgiova:
This PR is passing the tests: https://gitlab.com/Northern.tech/Mender/mender-helm/-/pipelines/1687722143 (changelog CI error is not related and it has been addressed in another PR)

oldgiova (approved):

PR #457, opened 2025-03-10
perf: Clean up unneeded Traefik middlewares

alfrunes:
A new path aliases in inventory was created before the v4.0.0 release. These middlewares are no longer required.


oldgiova (approved):

alfrunes:
There's an error in the inventory devices API router. Will have to postpone this one for the next major release.

PR #464, opened 2025-04-16
feat: add minReadySeconds

oldgiova:
minReadySeconds could be useful when a new deployment came, and you have to coinsider the pod ready some time after the readinessProbe

Ticket: MC-7895


alfrunes (commented):
I'm missing a `minReadySeconds` for `generate-delta-worker` StatefulSet. Otherwise I think it's a nice addition :+1:

alfrunes (approved):

PR #467, opened 2025-05-07
MC-6867 - feat: added Mender HTTP headers to display the chart version + name

mzedel:
as in: https://github.com/NorthernTechHQ/alvaldi-helm/pull/28

### DIFF
--- a/mender/templates/api-gateway/configmap.yaml
+++ b/mender/templates/api-gateway/configmap.yaml
@@ -506,6 +506,8 @@ data:
             browserXssFilter: true
             customRequestHeaders:
               "X-Forwarded-Proto": "{{ $scheme }}"
+              "X-Mender-Version": "{{ .Chart.Version }}"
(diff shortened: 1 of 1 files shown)


mender/templates/api-gateway/configmap.yaml
```diff
@@ -506,6 +506,8 @@ data:
             browserXssFilter: true
             customRequestHeaders:
               "X-Forwarded-Proto": "{{ $scheme }}"
+              "X-Mender-Version": "{{ .Chart.Version }}"
```
oldgiova, line 509: Great stuff! In Alvaldi the headers are lowercase, but I guess it's fine

oldgiova (approved):

mender/templates/api-gateway/configmap.yaml
mzedel: yep, was just aligning with the header above - it _shouldn't_ really matter at all:  https://datatracker.ietf.org/doc/html/rfc7230#section-3.2 but if we come across anything I'll lowercase right away 😄

PR #468, opened 2025-05-07
docs: update yaml customize path links

guspan-tanadi:


oldgiova (approved):
Thank you for your contribution!

PR #473, opened 2025-05-22
chore: upgrade helm chart for Mender 3.7.11 release

danielskinstad:
First time doing this, please make sure it's correct


oldgiova (changes requested):
Great job! Only you should amend the Chart.yaml updated version, because the release automation will take care of it.

Additionally, to make this work, you should amend your commit with `fix: ` so release please will bump the patch version.

Thank you!

oldgiova (approved):

PR #480, opened 2025-05-27
fix: display Mender Version

oldgiova:
Display the actual Mender version based on the computed Tag, not just on Chart.Appversion

Ticket: None


kjaskiewiczz (approved):
thanks!

PR #491, opened 2025-06-12
fix: some migration jobs don't run for open source

oldgiova:
The migration jobs for the Workflows and Useradm service are not running for the Open Source helm chart by default. This fix enables them.

Ticket: MC-8005


danielskinstad (approved):
Looks good to my limited knowledge

PR #492, opened 2025-06-17
chore: enable MEN-8168 feature flag in the useradm service

kjaskiewiczz:


mender/values.yaml
```diff
@@ -835,6 +835,7 @@ useradm:
     USERADM_REDIS_TIMEOUT_SEC: "1"
     USERADM_REDIS_LIMITS_EXPIRE_SEC: "3600"
     USERADM_TOTP_ISSUER: Mender
+    USERADM_MEN-8168: "true"
```
alfrunes: Unfortunately, this change won't introduce any change on the templates itself. All the `env.*` values are actually hard-coded values in the pod template. You could use `customEnvs` value, but I don't think that's good practice.
I think this is a better candidate to put under the `featureGates` value as this is something we plan on making the default (remove) on the next major release.

PR #507, opened 2025-07-02
api-gateway: add deployment annotations to template

chriswiggins:
We have a service that auto-reloads deployments when secrets/configmaps change (i.e cert-manager renews a certificate). See https://github.com/stakater/Reloader

We need to be able to set annotations on the api-gateway to support this, so I've added this into the template.

For completeness, if I add this to the api-gateway should I go through and add this option to all other k8s deployment templates?


chriswiggins:
Awesome - lets keep it minimal for now, can always add more in future? Have added README changes

PR #531, opened 2025-08-29
fix(generate-delta-worker): Set automigrate to false by default

alfrunes:


oldgiova:
> @alfrunes What does this fix fixes? I remember that without the automigrate: true, the initialization with an empty db could fail, but I'm not able to find any history ATM, and I could be wrong

Now I remember it: since there's no migration job for generate-delta-worker, the only way currently to initialize it is to set `automigrate: false`

PR #534, opened 2025-09-09
Add mountSecrets to `useradm`, `deviceauth` and `tenantadm` service

alfrunes:


mender/values.yaml
```diff
@@ -462,6 +462,20 @@ device_auth:
   # - name: LOG_LEVEL
   #   value: DEBUG
 
+  # Custom volumeMounts for mouting
```
oldgiova, line 465:
```suggestion
  # Custom volumeMounts for mounting
```

mender/values.yaml
```diff
@@ -875,6 +903,20 @@ useradm:
...
```
oldgiova, line 906:
```suggestion
  # Custom volumeMounts for mounting
```

oldgiova (approved):
Nice! Only very little nitpicks; approved

PR #536, opened 2025-09-19
Included dependency from bitnami to bitnamisecure

oldgiova:
By default this helm chart comes with convinient out-of-the-box dependencies from Bitnami: Redis and MongoDB. These dependencies are not supposed to be used in production.
The dependencies was moved from the legacy bitnami repository to the new bitnamisecure repository with only the latest tag available.

Please don't use these dependencies in production, but for testing purposes only.

Ticket: MC-8173

Testing pipeline: https://gitlab.com/Northern.tech/Mender/mender-helm/-/pipelines/2049106206


oldgiova:
We chose to include mongo and redis manifests, instead of using bitnamisecure.

PR #542, opened 2025-09-26
fix: increase default limits for delta generator

merlin-northern:
Changelog: Title
Ticket: MEN-8765

### DIFF
--- a/mender/values.yaml
+++ b/mender/values.yaml
@@ -546,11 +546,15 @@ generate_delta_worker:
       whenScaled: "Delete"
   resources:
     limits:
-      cpu: 100m
-      memory: 1024Mi
+      cpu: 2
+      # Warning: the artifact size together with xdelta3 arguments
+      # of window sizes may result in exhaustion of this memory limit,
+      # and the delta generation could fail or take forever to finish.
+      # Please tune it here and/or with xdelta3 arguments
+      memory: 4096Mi
     requests:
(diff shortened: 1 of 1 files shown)


mender/values.yaml
```diff
@@ -546,11 +546,11 @@ generate_delta_worker:
...
     limits:
-      cpu: 100m
-      memory: 1024Mi
+      cpu: 2
+      memory: 4096Mi
     requests:
-      cpu: 100m
-      memory: 128Mi
+      cpu: 1
+      memory: 2048Mi
```
oldgiova, line 553:
```suggestion
resources:
limits:
      cpu: 2
      # Warning: if your file size exceeds this memory limit,
      # the delta generation could fail.
      # Please tune it with:
      # <your maximum file size> + <your maximum file size> * 20%
      memory: 4096Mi
requests:
      cpu: 1
      memory: 2048Mi
```

merlin-northern: yeah good one. there is only one thing I have to do, read xdelta3 sources, because I bet we have to mention at least one window size (sine you can get crazy with the parameters to the generator, I have to check it allocates memory vs the values given)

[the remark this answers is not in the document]
merlin-northern: fixed.

merlin-northern: fixed/added. thanks.

PR #642, opened 2026-04-07
feat: added inventory addr to tenantadm pod template

frodeha:
**Description**
Added `INVENTORY_ADDR` to the tenantadm podtemplate as a result of the dependency introduced in https://github.com/mendersoftware/mender-server-enterprise/pull/1124

Ticket: MEN-9470


oldgiova (changes requested):
The change is good, but the commit message won't trigger a new release; please use either `fix` or `feat` (the second one I think is most in scope)

alfrunes (approved):

oldgiova (approved):

PR #664, opened 2026-05-08
Align prod with master

oldgiova:


alfrunes (approved):
LGTM, but shouldn't it be "Align prod with master"?

oldgiova:
> LGTM, but shouldn't it be "Align prod with master"?

Yeah of course :man_facepalming:

PR #672, opened 2026-05-25
Breaking: Update NATS helm subchart

oldgiova:
* Update NATS Helm subchart from 0.19.17 to 2.14.0
* move api gateway secrets to kubernetes standard
* remove deprecated device_license_count

Co-authored-with: Claude


mender/Chart.yaml
alfrunes, line 1: Shouldn't we also bump the major version here? There are already breaking changes in the `values.yaml` along with migration instructions. Or are we relying on the release tools to do it?

mender/values.yaml
```diff
@@ -9,7 +9,9 @@ global:
     existingSecret: ""
     URL: ""
   nats:
+    # DEPRECATED: use nats.existingSecret instead. Will be removed in the next major release.
     existingSecret: ""
+    # DEPRECATED: use nats.URL instead. Will be removed in the next major release.
```
alfrunes, line 14: But isn't this bumping the major release? Won't these options disappear?

UPGRADE_from_v7_to_v8.md
```diff
@@ -0,0 +1,201 @@
...
+  cronjobs:
+    enabled: false
+```
+
+## Deprecated values (still work in 8.x, removed in 9.x)
+
+| Deprecated key | Use instead |
+|---|---|
+| `global.nats.existingSecret` | `nats.existingSecret` |
+| `global.nats.URL` | `nats.URL` |
```
alfrunes, line 193: Ohh, got it. We're deferring removal until 9.0.

alfrunes: On the other hand, since the clusters cannot be updated in-place, should we just rip the bandaid and enforce this change?

alfrunes (approved):
LGTM, but there are so many other breaking sub-chart value changes, that I'm wondering if we should just do the same for the two `global`s.

mender/Chart.yaml
oldgiova: We have to rely on release please, otherwise it would bump another version

mender/values.yaml
oldgiova: I thought this was just the announcement for it, then we'll remove the global.nats entry on the next major update.
This breaking change is mainly driven by NATS major update, and we're breaking the changes announced already (the featuregates), but I don't think we can introduce other breakages not announced

UPGRADE_from_v7_to_v8.md
oldgiova: > On the other hand, since the clusters cannot be updated in-place, should we just rip the bandaid and enforce this change?

Well, I think we could... Ok, let's do it!

mender/values.yaml
alfrunes: But strictly speaking, aren't there many breakages under the  `nats.*` values already?

oldgiova: Follow up here: https://github.com/mendersoftware/mender-helm/pull/672/changes#r3311184932

I think you have right, we could remove these options with this release already

UPGRADE_from_v7_to_v8.md
oldgiova: I thought a bit more about it... a customer using an external NATS cannot upgrade in place because of missing grace period for deprecated changes:

Before v8:
```
global:
  nats:
    existingSecret: my-secret

nats:
  enabled: false
```

With v8:
```
global:
  nats:
    existingSecret: my-secret # <-- this causes a failure

nats:
  existingSecret: my-secret
```

So we actually need a grace period where both of the values can coexists to have  a smooth upgrade

oldgiova:
@alfrunes I added these commits regarding findings during the review:
* https://github.com/mendersoftware/mender-helm/pull/672/changes/7f371ee1175b8d4c6faa7fc55f5926bd3ed8b1f7
* https://github.com/mendersoftware/mender-helm/pull/672/changes/f7483bd6adb9faad70f344a05983065099bd58a3
* https://github.com/mendersoftware/mender-helm/pull/672/changes/867b2a493cb659d8bd25d5c8d34cf3819812d1fb
* https://github.com/mendersoftware/mender-helm/pull/672/changes/a4eb1952ef19c6f2007657dc747e06ce44b83b9f

Also rebased on top of master.

UPGRADE_from_v7_to_v8.md
alfrunes: You're right! Let's do as you suggest and defer the breaking change.

alfrunes (approved):

oldgiova:
Rebased

PR #685, opened 2026-05-31
Add support script

LudvigAnderson:
**NOTE:** there are 2 commits here. The first simply adds the support script as it is in [Roberto's repo](https://github.com/oldgiova/mender-server-support-tool/blob/main/mender_server_support_tool.sh). The second commit actually contains the changes to this repo (and the support script!). I did this so that you will be able to more easily see what changes I made to the support script by looking at the 2nd commit.

Some more things to note:
* There are now 2 ways to run the script:
  * It can still be run manually just as before. I tested manually that it works.
  * It can be run by creating a job, e.g.: `kubectl create job --from=cronjobs/mender-support mender-support-test`. I tested this manually by installing this with Helm on a local k3s cluster. It worked too.
* I use the public `alpine/k8s` image for the job, but we should maybe have our own image to use instead.
* I used Claude to search through the script for any `kubectl` commands and put the appropriate permissions in `mender/templates/support/rbac.yaml`.
* If running the script by creating a job, wait for it to finish, and then run the command that is at the bottom of the logs. When running for example `kubectl logs -f jobs/mender-support-test`, it will say something like this at the bottom of the logs:
> Save the bundle locally by running:
  `kubectl logs -n default mender-support-test-4xtp5 | awk '$0=="-----BEGIN MENDER SUPPORT BUNDLE-----"{f=1;next} $0=="-----END MENDER SUPPORT BUNDLE-----"{f=0} f' | base64 -d > mender_support_bundle.tar.gz`
Then review the contents, transfer over an encrypted channel, and shred when done.

### DIFF
--- a/mender/files/mender_server_support_tool.sh
+++ b/mender/files/mender_server_support_tool.sh
@@ -4,14 +4,23 @@ set -euo pipefail
 
 # Global variables
 TMP_DIR=""
-NAMESPACE=""
-HELM_RELEASE=""
+NAMESPACE="${NAMESPACE:-}"
+HELM_RELEASE="${HELM_RELEASE:-}"
 TIMESTAMP=$(date +"%Y%m%d_%H%M%S")
 RAND_SUFFIX=$(od -An -N3 -tx1 /dev/urandom 2>/dev/null | tr -d ' \n')
 SUPPORT_BUNDLE="mender_support_${TIMESTAMP}_${RAND_SUFFIX}.tar.gz"
 MASK_SECRETS="${MASK_SECRETS:-true}" # Can be disabled with MASK_SECRETS=false
 MAX_LOG_LINES="${MAX_LOG_LINES:-1000}"
 MONGO_IMAGE="${MONGO_IMAGE:-mongo:8.0}"
+
+# Detect whether we have an interactive terminal. When false (e.g. running inside
+# a Kubernetes Job), prompts are skipped and NAMESPACE / HELM_RELEASE must be
+# provided via the environment.
+if [[ -t 0 ]]; then
+  INTERACTIVE=true
+else
+  INTERACTIVE=false
+fi
 
 # Colors for output
 RED='\033[0;31m'
@@ -65,10 +74,14 @@ check_command() {
     ;;
   *)
     print_msg "$YELLOW" "⚠ Warning: $cmd is installed in non-standard location: $cmd_path"
-    read -p "Do you want to continue? (y/N): " -n 1 -r
-    echo
-    if [[ ! $REPLY =~ ^[Yy]$ ]]; then
-      return 1
+    if [[ "$INTERACTIVE" == true ]]; then
+      read -p "Do you want to continue? (y/N): " -n 1 -r
+      echo
+      if [[ ! $REPLY =~ ^[Yy]$ ]]; then
+        return 1
+      fi
+    else
+      print_msg "$YELLOW" "  Non-interactive mode: continuing despite non-standard location"
     fi
     ;;
   esac
@@ -91,9 +104,12 @@ check_requirements() {
     exit 1
   fi
 
-  # Verify kubectl cluster connectivity before proceeding
+  # Verify kubectl cluster connectivity before proceeding.
+  # `kubectl version` contacts the API server (proving reachability) without
+  # requiring access to kube-system, so it works under least-privilege RBAC.
+  # `kubectl cluster-info` is avoided as it lists kube-system services we cannot read.
   print_msg "$YELLOW" "Verifying cluster connectivity..."
-  if ! kubectl cluster-info >/dev/null 2>&1; then
+  if ! kubectl version >/dev/null 2>&1; then
     print_msg "$RED" "Error: Cannot connect to Kubernetes cluster."
     print_msg "$RED" "Check your kubeconfig and cluster availability."
     exit 1
@@ -123,6 +139,19 @@ validate_name() {
 
 # Function to select namespace
 select_namespace() {
+  # If a namespace was provided via the environment, use it as-is (no prompt).
+  if [[ -n "$NAMESPACE" ]]; then
+    validate_name "$NAMESPACE" "namespace"
+    print_msg "$GREEN" "Using namespace from environment: $NAMESPACE"
+    echo ""
+    return
+  fi
+
+  if [[ "$INTERACTIVE" != true ]]; then
+    print_msg "$RED" "Error: NAMESPACE must be set when running non-interactively"
+    exit 1
+  fi
+
   print_msg "$YELLOW" "Fetching available namespaces..."
 
   # Get namespaces and store in array (safe from word splitting)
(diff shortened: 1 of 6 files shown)


LudvigAnderson:
Go here to see the changes to the support script: https://github.com/mendersoftware/mender-helm/pull/685/changes/31013c122c98ee1163fed8a1bcd528156d9f5585#diff-03c7437f4b542fcb0e64f20bc4a94ca86b46fa8f7aea02f27e01b6368e04b612

oldgiova (commented):
I think this is a nice addition. I'm jus wondering if we can introduce some kind of tests to prevent secrets leaking in the script, what do you think?

Anyway, the second commit header is the same as the first one, could you please change it to be more meaningful? Additionally, the `feat` type would trigger a new Mender Helm Chart major release, which is not what we want here, maybe a `chore` would fit better?

oldgiova (changes requested):
I don't think putting it into a cronjob could be viable, because to activate it you need a new deploy, and most of the time is not what a customer want (or maybe the cluster is already broken and you cannot change it).
I'd rather use the script as intended: you download it and you run it from your workstation. But it could stay in this repo, as this is public

mender/files/mender_server_support_tool.sh
```diff
@@ -0,0 +1,1206 @@
...
+    -e 's/(-----BEGIN[^-]+-----)[^-]+(-----END[^-]+-----)/\1***CERTIFICATE_CONTENT***\2/g' \
+    -e 's/(Bearer[[:space:]]+)[^[:space:]]+/\1***MASKED***/gi' \
+    -e 's/(Basic[[:space:]]+)[^[:space:]]+/\1***MASKED***/gi' \
+    -e 's/(ghp_[a-zA-Z0-9]{36})/***GITHUB_TOKEN***/g' \
+    -e 's/(ghs_[a-zA-Z0-9]{36})/***GITHUB_TOKEN***/g' \
+    -e 's/(glpat-[a-zA-Z0-9_-]{20,})/***GITLAB_TOKEN***/g' \
+    -e 's/(sk_live_[a-zA-Z0-9]{24,})/***STRIPE_LIVE_KEY***/g' \
+    -e 's/(sk_test_[a-zA-Z0-9]{24,})/***STRIPE_TEST_KEY***/g' \
+    -e 's/(xox[baprs]-[a-zA-Z0-9-]+)/***SLACK_TOKEN***/g'
+}
```
alfrunes, line 382: This looks incredibly brittle and should not be relied upon for masking secrets.

mender/files/mender_server_support_tool.sh
```diff
@@ -0,0 +1,1206 @@
...
+    print('-----------------------------------------');
+    print('Processing database: ' + database.name);
+    print('-----------------------------------------');
+    var currentDb = db.getSiblingDB(database.name);
+    currentDb.getCollectionNames().forEach(function(collectionName) {
+      print('Collection: ' + collectionName);
+      currentDb[collectionName].find().forEach(function(doc) { printjson(doc); });
+    });
+  }
+});
```
alfrunes, line 906: This database scheme for useradm was abandoned several years ago.

mender/files/mender_server_support_tool.sh
```diff
@@ -0,0 +1,1206 @@
...
+quit;
```
alfrunes, line 996: Same as useradm

mender/files/mender_server_support_tool.sh
```diff
@@ -0,0 +1,1206 @@
...
+if (dbNames.indexOf('inventory') !== -1) {
+  print('-----------------------------------------');
+  print('Processing database: inventory');
+  print('-----------------------------------------');
+  var inventoryDb = db.getSiblingDB('inventory');
+  inventoryDb.getCollectionNames().forEach(function(collectionName) {
+    print('Collection: ' + collectionName);
+    inventoryDb[collectionName].find().forEach(function(doc) { printjson(doc); });
+  });
+}
```
alfrunes, line 1030: This one's empty if `enterprise == true`

mender/files/mender_server_support_tool.sh
```diff
@@ -0,0 +1,1206 @@
...
+  js_script=$(cat <<'JSEOF'
+use tenantadm;
+var collections = db.getCollectionNames();
+collections.forEach(function(collectionName) {
+  print('Collection: ' + collectionName);
+  db[collectionName].find().forEach(function(doc) {
+    printjson(doc);
+  });
+});
+quit;
```
alfrunes, line 1074: This one doesn't exist for open source.

mender/files/mender_server_support_tool.sh
```diff
@@ -0,0 +1,1206 @@
...
+  # Collect information
+  collect_helm_history
+  collect_helm_list
+  collect_helm_values
+  collect_pods
+  collect_pod_logs
+  collect_configmaps
+  collect_secrets_list
+  list_tenants
+  mongo_dump_useradm
```
alfrunes, line 1176: Could use mongodump for these to make it easier to restore in a test environment?
```suggestion
  mongodump --db=useradm --gzip --archive=useradm.bson.gz
```

alfrunes (commented):
Although I can see this as a useful tool for debugging, I'm a bit concerned with maintaining this along side the application.
It is making many assumptions, and some of them don't hold across open source and enterprise. As it's not tied into the application it may also break at any point in time. Moreover, I'm a concerned of having this in the helm chart as I wouldn't run this in any production environment by accident. And as @oldgiova already mentioned, you probably want to run this as a one-off Job and not a recurring CronJob.

mender/files/mender_server_support_tool.sh
oldgiova: Actually, this was introduced recently to support a customer with Mender Server 3.2: https://github.com/oldgiova/mender-server-support-tool/commit/208546e3729ba2f4a028ca8b61d890e699d97664

Anyway, that was a kind of self-experiment, I agree to streamline the script more, if we want to maintain it officially

LudvigAnderson:
@oldgiova @alfrunes , I have read your comments. Thanks for feedback!

> you probably want to run this as a one-off Job and not a recurring CronJob.

It's not a recurring CronJob. The CronJob is suspended and with a schedule of every 31st of February, which will never happen. It's just there to package the script so that the user can run `kubectl create job --from=cronjobs/mender-support <job-name>` without necessarily needing the script locally. If you are opposed to this, that's okay, but it is does indeed function as a one-off Job.

> I don't think putting it into a cronjob could be viable, [...] I'd rather use the script as intended: you download it and you run it

That's fair, and we will always be able to send them the script or point them to it if they need it. But isn't there one advantage with shipping it with mender-helm? If for example a new customer deploys 4.1 and then in the future when we have 4.2, the script breaks, so it needs to be fixed. If so, maybe the new script is incompatible with 4.1, but the one which is already shipped with mender-helm is still compatible. I guess a different way to deal with this is to create a `mender-server-support-script-v4-1.sh` and then just create a new `mender-server-support-script-v4-2.sh` when the time comes. Please correct me if I have any wrong assumptions:)

> It is making many assumptions, and some of them don't hold across open source and enterprise. As it's not tied into the application it may also break at any point in time.

I forgot that this repo is public :laughing:. Well, do you think it's better to rewrite this script in Go as part of `mender-server-enterprise` so that it can be part of CI and whatnot? What are your thoughts?:)

LudvigAnderson:
Closing after server team meeting discussion.

PR #689, opened 2026-06-01
Align Prod from master

oldgiova:


mender/CHANGELOG.md
```diff
@@ -1126,6 +1126,22 @@
...
+* **generate-delta-worker:** removed wrong maxSurge option ([7db25b2](https://github.com/mendersoftware/mender-helm/commit/7db25b2d2e65bcc22dbcaba66dca87b1ee0f49af))
+* **generate-delta-worker:** removed wrong maxSurge option ([c7bc9bf](https://github.com/mendersoftware/mender-helm/commit/c7bc9bfe30a7c57d1751c92481c41267fdf86ba8))
+
+## [7.7.3](https://github.com/mendersoftware/mender-helm/compare/mender-7.7.2...mender-7.7.3) (2026-05-26)
+
+
+### Bug Fixes
+
+* remove invalid progressDeadlineSeconds from statefulset ([48a5a60](https://github.com/mendersoftware/mender-helm/commit/48a5a60ac913493ee0f03ed966e8a1d2605538e1))
+* remove invalid progressDeadlineSeconds from statefulset ([1f814bb](https://github.com/mendersoftware/mender-helm/commit/1f814bbdf782132c7e427ca3652aa53f2d7ee5b5))
```
alfrunes, line 1143: There seems to be some duplicate changelog entries here, one of them comes from the merge commit. I think we need to adjust the configuration to exclude merge commits.

alfrunes (approved):

mender/CHANGELOG.md
oldgiova: Well, we should squash merge to avoid it: https://github.com/googleapis/release-please/issues/2476

alfrunes: But isn't it using git-cliff(1) to generate this changelog? I thought the [`filter_unconventional`](https://github.com/mendersoftware/mendertesting/blob/a192813e86a44e005501f8da779360beee6d6f7a/utils/cliff.toml#L89) option should filter the merge commits since they're not conventionally formatted.

oldgiova: Right, but git-cliff has the same behavior: still including the merge commit.  

Filter_unconventional we already tried and it has some unwanted side effects if I good remember.

I think we need a dedicated ticket and a careful evaluation for this

PR #705, opened 2026-06-16
ci: fix CloudFormation stack deletion failures on EKS cleanup

oldgiova:
The cleanup jobs were leaving orphaned CloudFormation stacks because:
- The ecr-pull-through-cache inline IAM policy added to the node role during cluster setup was never removed before eksctl delete cluster, causing CloudFormation to fail deleting the node IAM role.
- The AWS Load Balancer Controller provisioned ELBs outside CloudFormation; deleting the cluster without first removing Helm releases left those ELBs attached to the VPC subnets, blocking VPC deletion.
- allow_failure: true on the cleanup jobs silently absorbed all failures.

Fix by adding three new YAML anchors:
- delete_ecr_inline_policy: removes the inline policy via aws iam before the cluster is torn down.
- pre_eks_delete_cleanup: iterates all namespaces, deletes all Helm releases, and waits 60s for ELB deprovisioning before eksctl runs.
- post_test_helm_cleanup: replaces the per-job after_script that only deleted the mender release; now deletes all releases in the namespace (seaweedfs, mender-mongo, nats, and mender).

Replace allow_failure: true with an explicit eksctl get cluster existence check so cleanup failures are surfaced rather than swallowed.

Co-Authored-with: Claude

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -126,9 +126,15 @@ renovate:
     --output text))
   - aws sts get-caller-identity
 
+.post_test_helm_cleanup: &post_test_helm_cleanup
+  - |
+    echo "INFO - cleaning up all Helm releases in namespace ${NAMESPACE}"
+    helm list -n "${NAMESPACE}" -q 2>/dev/null \
+      | xargs -r helm delete -n "${NAMESPACE}" --timeout 5m || true
+
 .pre_helm_tests: &pre_helm_tests
   - eksctl utils write-kubeconfig --cluster=${EKS_CLUSTER_NAME}
-  - kubectl create ns "${NAMESPACE}"
+  - kubectl create ns "${NAMESPACE}" --dry-run=client -o yaml | kubectl apply -f -
   - kubectl config set-context --current --namespace="${NAMESPACE}"
   - echo "INFO - installing helm from scratch"
   - aws ecr get-login-password --region eu-central-1 | helm registry login --username AWS --password-stdin ${ECR_REGISTRY}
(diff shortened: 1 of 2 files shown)


.gitlab-ci.yml
```diff
@@ -126,9 +126,125 @@ renovate:
...
+.post_test_helm_cleanup: &post_test_helm_cleanup
+  - |
+    echo "INFO - cleaning up all Helm releases in namespace ${NAMESPACE}"
+    helm list -n "${NAMESPACE}" -q 2>/dev/null \
+      | xargs -r helm delete -n "${NAMESPACE}" --timeout 5m || true
+
 .pre_helm_tests: &pre_helm_tests
   - eksctl utils write-kubeconfig --cluster=${EKS_CLUSTER_NAME}
-  - kubectl create ns "${NAMESPACE}"
+  - kubectl create ns "${NAMESPACE}" --dry-run=client -o yaml | kubectl apply -f -
```
frodeha, line 247: What's the difference here? Isn't this dry-run + pipe just a more cumbersome way to create the namespace?

.gitlab-ci.yml
```diff
@@ -760,15 +870,23 @@ publish:helm_chart_publishing:
...
       aud: https://gitlab.com
   before_script:
     - *set_eks_helmci_vars
   script:
-    - echo "INFO - deleting the temporary EKS cluster"
-    - eksctl delete cluster -n ${EKS_CLUSTER_NAME}
+    - *delete_ecr_inline_policy
+    - *pre_eks_delete_cleanup
+    - |
+      echo "INFO - checking if cluster ${EKS_CLUSTER_NAME} exists before deletion"
```
frodeha, line 883: I'm not sure so I'm just asking, but if the cluster doesn't exist, will the `pre_eks_delete_cleanup` produce an error failing the job? Should we if so check cluster existence before starting that cleanup?

frodeha (approved):
LGTM - I only have some questions mostly to solidify my own understanding

.gitlab-ci.yml
frodeha: Ahh, I see now in the commit message:
```
kubectl create ns and kubectl create secret both fail with
AlreadyExists when a job is retried while the cluster is still
up. Switch both to --dry-run=client -o yaml | kubectl apply -f -
so re-runs are safe regardless of prior state.
```

.gitlab-ci.yml
oldgiova: In a way, yes, but all the steps with this yaml anchor are optionals (ending with `|| true` or `/dev/null` ) because every single step might found already-deleted resources. So actually this script can't fail, neither if the EKS cluster is already destroyed by another .post job or manually

PR #731, opened 2026-07-23
feat: update NoAuth rule to include users/me/email-change/complete

kjaskiewiczz:


kjaskiewiczz:
Should we do it this way or should we change https://docs.mender.io/api/complete-email-change to include `auth` (it's already public if I'm not wrong, but I'm pretty sure nobody started using it yet, since we don't have UI)?
What do you think @alfrunes @bahaa-ghazal?

bahaa-ghazal:
if we change it to `auth` we need to change the UI path for completion and the workflow and many things. I would personally just go with this fix because it is simpler

bahaa-ghazal (approved):

alfrunes (approved):
It does complicate the routing a little, but it's one or the other...

PR #733, opened 2026-07-24
fix: configmaps not created for resources created by helm hooks

alfrunes:


merlin-northern (approved):

merlin-northern:
wow. I have no idea how that could have happened. also I am not an expert here.

oldgiova:
I'm not sure this is going to work, because it opens a window for deleting the regular configmap from helm hooks; I would let alone regular configmaps from helm hooks.
Maybe if the issue is for deviceconnect only, it's better to extend there the same `-prerelease` configmap as well as deployment service: https://github.com/mendersoftware/mender-helm/blob/master/mender/templates/deployments/configmap-prerelease.yaml ?
@alfrunes

alfrunes:
Ohh, seems like I was a bit too fast! Let me create a new PR...

PR #743, opened 2026-07-31
feat: default USERADM_JWT_ISSUER to global.url FQDN

lluiscampos:
Make the useradm chart fall back to the bare FQDN derived from global.url (analog to USERADM_BASE_URL). The scheme is stripped away.


alfrunes (approved):
Thank you :pray: That's a much more meaningful default!

PR #749, opened 2026-08-10
ci: migrate mender-helm fully to renovate

p-targowicz:
Part of QA-1485

Dropped dependabot. It was bumping the mongo redis and traefik image tags in mender/values.yaml which renovate's helm-values manager covers instead

Renamed renovate.json to renovate.json5 so the config can carry comments. Checked there is no RENOVATE_CONFIG_FILE variable at project or group level so nothing breaks on the rename. Also migrated the deprecated regexManagers and fileMatch spelling

Kept the existing policy that mongo nats and redis only get patch bumps. Raised prConcurrentLimit from 1 to 5 since renovate now covers what dependabot did

Switched from the explicit sre reviewers to reviewersFromCodeOwners which is what the dependabot image bumps already used - CODEOWNERS maps /mender/ to sre backend and frontend

Left the CI wiring alone - renovate already runs here via the upstream renovate-runner component rather than the mendertesting wrapper


renovate.json5
```diff
@@ -0,0 +1,79 @@
...
+  ],
+
+  // CODEOWNERS drives reviewer assignment, which is what the Dependabot image bumps
+  // on mender/values.yaml already used - keep its entries current.
+  "reviewersFromCodeOwners": true,
+
+  // chore(deps): <desc>. Dependabot used a "fix" prefix here, which put dependency
+  // bumps in the wrong changelog section.
+  "semanticCommits": "enabled",
+  "semanticCommitType": "chore",
```
oldgiova: We have release-please with this repo; this means it could skip a `chore(deps)` and don't trigger a new release.

I would scope `chore(deps)` for the CI changes (like .gitlab-ci.yml) which are not triggering a new release anyway, and for the other changes I would set `fix(deps)`.

See this proposal for more context: https://github.com/mendersoftware/mender-qa/pull/891/changes

p-targowicz: changed to https://github.com/mendersoftware/mender-qa/pull/891/changes

oldgiova (approved):

PR #771, opened 2026-09-12
feat: add a dedicated rate limit for the artifact list endpoint

oldgiova:
GET /api/management/v[0-9a-z]+/deployments/artifacts returns a large response body and is cheap to poll. The general per-IP ratelimit middleware counts requests, not bytes, so at its configured 50 req/s it authorises roughly 6.4 MB/s from a single client on this route.

Add an optional artifactsListRateLimit middleware and a dedicated router for that path, following the existing authRateLimit pattern. The router takes priority 100 so it wins over deploymentsMgmt, which matches the same prefix by rule length.

The rule is anchored, so artifacts/list, artifacts/{id} and the POST upload path are unaffected and keep the general ratelimit.

Both blocks are rendered only when api_gateway.artifactsListRateLimit is set, so the chart is unchanged for anyone who does not configure it.

Ticket: MC-8877

Co-authored-with: Claude


oldgiova:
Moved to the useradm internal rate limits

